PR #1, opened 2019-03-12
Add incomplete_address? check method

JuanCrg90:
Solidus has deprecated the `empty?` method on https://github.com/solidusio/solidus/pull/1686 . 
This adds a private `incomplete_address?` method for avoid future problems and remove the deprecation warning in projects that uses this gem.

### DIFF
--- a/lib/super_good/solidus_taxjar/tax_calculator.rb
+++ b/lib/super_good/solidus_taxjar/tax_calculator.rb
@@ -12,7 +12,7 @@ module SuperGood
 
       def calculate
         return no_tax if SuperGood::SolidusTaxJar.test_mode
-        return no_tax if order.tax_address.empty? || order.line_items.none?
+        return no_tax if incomplete_address?(order.tax_address) || order.line_items.none?
         return no_tax unless taxable_address? order.tax_address
 
         cache do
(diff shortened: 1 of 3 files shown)


lib/super_good/solidus_taxjar/tax_calculator.rb
```diff
@@ -150,6 +150,12 @@ def shipping_tax_label(shipment, shipping_tax)
       def line_item_tax_label(taxjar_line_item, spree_line_item)
         SuperGood::SolidusTaxJar.line_item_tax_label_maker.(taxjar_line_item, spree_line_item)
       end
+
+      def empty?(tax_address)
+        tax_address
+          .attributes
+          .except('id', 'created_at', 'updated_at', 'country_id').all? { |_, v| v.nil? }
+      end
```
jarednorman: I'll accept this change as is, but I think we could definitely do something smarter here. What I think we actually care about is that [these attributes](https://github.com/SuperGoodSoft/solidus_taxjar/blob/master/lib/super_good/solidus_taxjar/api_params.rb#L16-L19) are present. What if we just checked those?

jarednorman: As much as I like the deduplication you get from relying on the `APIParams`, I think this code doesn't look very obvious; no one really should have to care about the structure of what comes back from the `APIParams` helper, except for the TaxJar API.

How about:

```ruby
def incomplete_address?(address)
  [
    tax_address.address1,
    tax_address.city,
    tax_address&.state&.abbr || tax_address.state_name,
    tax_address.zipcode,
    tax_address.country.iso
  ].any?(&:blank?)
end
```

jarednorman:
Thanks for the change! I'll cut a new release for you!

PR #2, opened 2019-03-13
Add guard clause for TaxLocation addresses

JuanCrg90:
From Solidus [Documentation](https://guides.solidus.io/developers/taxation/overview.html#use-code-spree-taxlocation-code-as-the-tax-address): 
> An order's tax_address can – through duck typing  – be a Spree::TaxLocation instead of the shipping address.

This adds a guard clause that validates if the tax_address is a TaxLocation, in this case we know that the address is incomplete.

### DIFF
--- a/lib/super_good/solidus_taxjar/tax_calculator.rb
+++ b/lib/super_good/solidus_taxjar/tax_calculator.rb
@@ -152,6 +152,8 @@ module SuperGood
       end
 
       def incomplete_address?(tax_address)
+        return true if tax_address.is_a?(Spree::Tax::TaxLocation)
+
         [
           tax_address.address1,
           tax_address.city,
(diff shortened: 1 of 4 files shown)


lib/super_good/solidus_taxjar/tax_calculator.rb
```diff
@@ -152,6 +152,8 @@ def line_item_tax_label(taxjar_line_item, spree_line_item)
       end
 
       def incomplete_address?(tax_address)
+        return true if tax_address.class == Spree::Tax::TaxLocation
```
jarednorman: What about `return true if tax_address.is_a?(Spree::Tax::TaxLocation`?

JuanCrg90: Updated

jarednorman:
Thanks @JuanCrg90!

PR #18, opened 2020-01-22
Remove unnecessary errors when validating address without country

spaghetticode:
When validating addresses, it may happen that the address has no country, so trying to fetch the country ISO raises an unnecessary error. Using ruby safe navigation prevents this to happen. 

Of course, the address will still result invalid eventually.

This has the extra benefit of avoiding unnecessary error reports to services such as Honeybadger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## master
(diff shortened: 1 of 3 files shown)


jarednorman:
Thank you @spaghetticode!

PR #26, opened 2020-10-09
Update extension name & add missing dependency

cmhnk:
These changes were necessary for me to be able to run `bundle exec rake sandbox` and `bundle exec rspec` successfully.

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -25,7 +25,7 @@ else
   BRANCH="master"
 fi
 
-extension_name="solidus_taxjar"
+extension_name="super_good-solidus_taxjar"
 
 # Stay away from the bundler env of the containing extension.
 function unbundled {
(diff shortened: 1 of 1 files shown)


bin/sandbox
```diff
@@ -53,6 +53,7 @@ gem 'solidus', github: 'solidusio/solidus', branch: '$BRANCH'
 gem 'solidus_auth_devise', '>= 2.1.0'
 gem 'rails-i18n'
 gem 'solidus_i18n'
+gem 'solidus_paypal_commerce_platform'
```
jarednorman, line 56: I don't think this _should_ be necessary, but I'll have to look into it. I think there's been some work to fix some issues that cropped up from it being the new default, but I don't remember the state of the specifically off the top of my head.

Can you make this PR just the extension name fix, and then I'll look into the paypal thing? (You can of course still keep using this fix locally.)

jarednorman (changes requested):

bin/sandbox
cmhnk: You got it! Going to open a new PR because if I force push to this branch, the name of the branch will (misleadingly) still have `and_add_missing_dependency` in it.

PR #29, opened 2021-01-13
Allow price to be configurable via a line_item method. 

AdnanTheExcellent:
This allows stores that dont follow the traditional price / quantity = amount calculation to set their own unit_price.


AdnanTheExcellent:
please ignore this, thought i was PRing into my fork but it PRed here instead

PR #31, opened 2021-01-28
Taxjar customer integration

AdnanTheExcellent:
This PR adds Taxjar Customer integration for tax exempt users. This allows for tax exemption information to be created within Solidus and sent to Taxjar via an api call. 

this PR adds:

ActiveRecord Models:
`SuperGood::SolidusTaxjar::Customer`
`SuperGood::SolidusTaxjar::ExemptRegion`

taxjar API calls:
`create_customer`
`update_customer`
`delete_customer`
`nexus_regions`

Breaking changes:
rename gem from `super_good-solidus_taxjar` to `super_good_solidus_taxjar`
rename `SolidusTaxJar` namespace to `SolidusTaxjar`
rename `API` to `Api` in a couple models

a migration to create the models + an install generator is included. Tests have been added for the new functionality. 

TODOs:

- [ ] Update version number
- [ ] Update readme
- [ ] Update Changelog


lib/spree/models/line_item_decorator.rb
```diff
@@ -0,0 +1,11 @@
...
+  module LineItemDecorator
+    protected
+
+    def unit_price
+      price
+    end
+
+    ::Spree::LineItem.prepend self
+  end
+end
```
jarednorman: Obviously just a tiny nit pick, but if you leave newlines at the end of files then GitHub won't leave these icons saying they're missing. I have my editor configured to do this automatically.

lib/spree/models/line_item_decorator.rb
```diff
@@ -0,0 +1,11 @@
+module Spree
+  module LineItemDecorator
+    protected
```
jarednorman: I don't see a reason this shouldn't be `private`.

lib/spree/models/line_item_decorator.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    def unit_price
+      price
+    end
```
jarednorman: I'd like to avoid decorating models unnecessarily when we could use configuration extension points.

Rather than add a method to the class that others might need to override, we could so something similar to other configurable functionality we already have; we could add a `line_item_unit_price_calculator` option to [the configuration](https://github.com/SuperGoodSoft/solidus_taxjar/blob/master/lib/super_good/solidus_taxjar.rb#L14-L53):

```ruby
attr_accessor :line_item_unit_price_calculator

# ...

self.line_item_unit_price_calculator = ->(spree_line_item) { spree_line_item.price }
```

app/decorators/models/super_good/solidus_taxjar/spree/user_decorator.rb
```diff
@@ -0,0 +1,20 @@
...
+        def self.prepended(base)
+          base.has_one :taxjar_customer, class_name: "SuperGood::SolidusTaxjar::Customer", inverse_of: :user
+          base.has_many :taxjar_exempt_regions, class_name: "SuperGood::SolidusTaxjar::ExemptRegion", through: :taxjar_customer
+
+          base.accepts_nested_attributes_for :taxjar_customer
+        end
+
+        def foo
+          'bar'
+        end
```
jarednorman, line 14: 👀

jarednorman (changes requested):
This seems like a good approach overall. I left one comment about how to approach the logic that needs to be configurable.

I'd appreciate a quick pass to address some of the formatting, mostly just to clean up the weird inconsistent/weird indentation in some of the specs you introduced. Additionally, I try to aim to follow the same conventions as Solidus when it comes to [git hygiene and messages](https://chris.beams.io/posts/git-commit/) on this project, so if you were kind enough to clean that up with a rebase, I'd appreciate that too.

@AlistairNorman You're using this currently, I know. Do you have any thoughts here?

app/decorators/models/super_good/solidus_taxjar/spree/user_decorator.rb
AdnanTheExcellent: whoops, that was from testing

lib/spree/models/line_item_decorator.rb
AdnanTheExcellent: i'll change that

lib/spree/models/line_item_decorator.rb
AdnanTheExcellent: I'm all for that

AdnanTheExcellent:
> 
> 
> This seems like a good approach overall. I left one comment about how to approach the logic that needs to be configurable.
> 
> I'd appreciate a quick pass to address some of the formatting, mostly just to clean up the weird inconsistent/weird indentation in some of the specs you introduced. Additionally, I try to aim to follow the same conventions as Solidus when it comes to [git hygiene and messages](https://chris.beams.io/posts/git-commit/) on this project, so if you were kind enough to clean that up with a rebase, I'd appreciate that too.
> 
> @AlistairNorman You're using this currently, I know. Do you have any thoughts here?

I'll address the formatting. I usually use rubocop but I noticed it wasn't configured for this extension. when I ran rubocop on it previously I got ~ 3500 offenses , which were mostly about quotations, that were drowning out everything. 

would you prefer everything squashed into one commit?

jarednorman:
Yeah, I prefer to use [Standard](https://github.com/testdouble/standard) on my extensions. I don't think I have any automation set up for it, but if you run your stuff through standard that should be good enough.

As far as commits, I prefer nice small commits, so if it's not too much work, my preference would be to have separate commits for:
- changing the name of the gem
- updating the constant names
- introducing the new records and schema
- whatever else is in here that I forgot

AdnanTheExcellent:
see #32 for rebased PR

PR #37, opened 2021-02-08
Add basic UI for Taxjar settings in solidus admin

Noah-Silvera:
This commit adds a tab to the taxes section of the solidus admin settings. 
This provides the foundation for further configuration of Taxjar through the solidus admin.

Right now, the new settings tab does not display any content.

If the API key is not set, the settings tab should just show a descriptive message asking you to provide an API token. This is to conform with the Taxjar certification checklist requirement to hide the Taxjar settings until an API token is provided.

The descriptive messaging also provides links to signup for Taxjar, and a link to help article for acquiring an API token, to comply with certification requirements. 

Manual Tests
----

1. Set a taxjar API key
     * [x] Ensure that you see a blank page in the Taxjar settings tab
1. Don't set a taxjar API key
     * [x] Ensure you see a descriptive message telling you to provide an API key

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog
- [x] Run a sandbox app and verify taxes are being calculated

Screenshots
---

Screen Shot 2021-02-21 at 20 58 44PST

Screen Shot 2021-02-22 at 13 30 00PST

PR #39, opened 2021-02-17
Add taxjar rails engine

Noah-Silvera:
In order to have the this extension certified by Taxjar, we will need to add significant UI to the solidus backend.

A prerequisite to this is adding a rails engine for Taxjar so components like controllers can be added to the extension and loaded by solidus.

### DIFF
--- a/lib/super_good/engine.rb
+++ b/lib/super_good/engine.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module SuperGoodSolidusTaxjar
+  class Engine < Rails::Engine
+    isolate_namespace Spree
+    engine_name 'super_good-solidus_taxjar'
+  end
+end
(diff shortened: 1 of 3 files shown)


adammathys (approved):

lib/super_good/engine.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module SuperGoodSolidusTaxjar
+  class Engine < Rails::Engine
+    isolate_namespace Spree
+    engine_name 'super_good-solidus_taxjar'
+
+    config.autoload_paths += %W(#{config.root}/lib)
```
jarednorman, line 8: Is there a reason you're adding lib to autoloads? I don't think we should.

jarednorman (changes requested):

lib/super_good/engine.rb
Noah-Silvera: No, there's no good reason. We have removed this.

We've also removed the rspec configuration, as it can be added when it is needed, in #37

jarednorman (approved):

PR #40, opened 2021-02-23
Add pull request template

Noah-Silvera:
What is the goal of this PR?
---

Add a basic pull request template

---

N/A

Merge Checklist
---

- [ ] Try to open a pull request and see if the template is used post merge

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,19 @@
+What is the goal of this PR?
+---
+
+
+
+How do you manually test these changes? (if applicable)
+---
+
+1. Do a thing
+    * [ ] Assert a result
+
+Merge Checklist
+---
+
+- [ ] Run the manual tests
+- [ ] Update the changelog
+
+Screenshots
+---


README.md
```diff
@@ -65,7 +65,9 @@ The extension provides currently two high level `calculator` classes that wrap t
 
 ## Development
 
-After checking out the repo, run `bin/setup` to install dependencies. Then, run `rake spec` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
+Before installing dependencies, make sure your environment variables are setup. Our CI uses something like this: `SOLIDUS_BRANCH=v2.10 DB=postgresql RAILS_VERSION="~> 5.2.0`. These environment variables need to be set when installing dependencies and running specs, so set them before the command like this: `SOLIDUS_BRANCH=v2.10 DB=postgresql RAILS_VERSION="~> 5.2.0 bin/setup`
```
adammathys, line 68: These variables all have default values so there shouldn't be any need to define them unless you're looking to target a specific setup. (So the "need to be set" is a little misleading.)

adammathys: This change also probably shouldn't be bundled in with the pull-request template PR.

adammathys (changes requested):

README.md
Noah-Silvera: Ahh, I messed up again and didn't rebase correctly onto master.

>These variables all have default values so there shouldn't be any need to define them unless you're looking to target a specific setup. (So the "need to be set" is a little misleading.)

You actually do need to set them! Whatever the default values are don't work, if you just run `bin/setup` and `bin/rake` without any env. But maybe instead of adding this readme, we should figure out why the default values don't work.

adammathys (approved):

jarednorman (approved):

PR #32, opened 2021-01-29
Add Taxjar customers and exempt regions integration

AdnanTheExcellent:
This PR adds Taxjar Customer integration for tax exempt users. This allows for tax exemption information to be created within Solidus and sent to Taxjar via an api call. 

this PR adds:

ActiveRecord Models:
`SuperGood::SolidusTaxjar::Customer`
`SuperGood::SolidusTaxjar::ExemptRegion`

taxjar API calls:
`create_customer`
`update_customer`
`delete_customer`
`nexus_regions`

Breaking changes:
rename gem from `super_good-solidus_taxjar` to `super_good_solidus_taxjar`
rename `SolidusTaxJar` module to `SolidusTaxjar`
rename `API` to `Api` in a couple classes

a migration to create the models + an install generator is included. Tests have been added for the new functionality. 

TODOs:

- [x] Update version number
- [x] Update readme
- [x] Update Changelog

PR #43, opened 2021-03-03
Support zeitwerk loading (0.18.0)

Noah-Silvera:
What is the goal of this PR?
---

Support loading this extension with the zeitwork loader

This is a breaking change due the renaming of the module

---

1. Set up this gem in a rails project that uses zeitwork
    * [x] Ensure it loads properly and can be used to calculate taxes

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,17 @@ dist: trusty
 sudo: false
 language: ruby
 cache: bundler
+addons:
+  postgresql: "9.3"
 jobs:
   include:
     - rvm: 2.5.8
       env: SOLIDUS_BRANCH=v2.9 DB=postgresql RAILS_VERSION="~> 5.1.0"
     - rvm: 2.6.6
       env: SOLIDUS_BRANCH=v2.10 DB=postgresql RAILS_VERSION="~> 5.2.0"
+    - rvm: 2.6.6
+      env: SOLIDUS_BRANCH=v2.10 DB=postgresql RAILS_VERSION="~> 6.0.0"
+
 before_install:
   - gem update --system
   - gem install bundler
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,26 @@
 # Changelog
 
-## master
+## v0.18.0
 
 - [#21](https://github.com/SuperGoodSoft/solidus_taxjar/pull/21) Migrated project to `solidus_dev_support`
 - [#22](https://github.com/SuperGoodSoft/solidus_taxjar/pull/22) Added support for TaxJar address validation API through `SuperGood::SolidusTaxJar::Addresses` class
 - [#34](https://github.com/SuperGoodSoft/solidus_taxjar/pull/34) Include API version in request headers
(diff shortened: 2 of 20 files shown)


Noah-Silvera:
@AdnanTheExcellent @nvandoorn and I don't quite understand the requirement for `active_storage` is needed? What required the dependency in your original PR #32 ? Thanks :)

AdnanTheExcellent:
@Noah-Silvera attaching the tax exemption documents. each exempt region has an attached tax exemption form. 

our tax exempt customer flow is:

1. The user creates a tax exemption (taxjar customer) with many exempt regions in their account settings. they submit their tax exempt forms and submit it to us.
2. Customer service (solidus admin) sees the requests, checks the tax exempt forms, and approves / rejects them in Solidus admin. Then the tax exempt forms are handled as necessary (outside scope of this extension), such as uploading them in taxjar admin or filing them somewhere else outside of solidus.

Noah-Silvera:
@AdnanTheExcellent that makes sense! We will let you make that breaking change in your PR then.

Noah-Silvera:
@AdnanTheExcellent  and @jarednorman we decided to make the upgrade to the zeitwork loaders **not** a 1.0.0 release as @nvandoorn brought up an excellent point that breaking changes as minor version changes when the version is below 1.X are acceptable in a lot of projects.

We envision a couple more breaking changes we will need to make in the near future (for example, @AdnanTheExcellent 's work in #32), and we don't want to end up at version 6.X within the next 6 months.

adammathys (approved):

jarednorman:
@Noah-Silvera I'm not sure if I put this anywhere, but I'd like to see if there's a way to resolve the migration issue that @AdnanTheExcellent identified without renaming the extension. If this is the only reasonable option, that's fine, but if there's a way to stick with the conventional naming, then I'd prefer to do that.

Noah-Silvera:
@jarednorman we gave it a test and the gem doesn't seem to need a rename in order to support rails 6.
@AdnanTheExcellent unless we've missed a reason we need the gem rename, i've dropped the commit.

jarednorman:
Sorry, that conversation was in DMs, here's the explanation from Adnan:

> the name `super_good-solidus_taxjar` breaks some things like install generators installing migrations due to the hyphen in the name. the migrations ported over will have `.super_good-solidus_tax_jar` appended to it which gives a `ActiveRecord::IllegalMigrationNameError` from the hyphen.

Noah-Silvera:
> Sorry, that conversation was in DMs, here's the explanation from Adnan:
> 
> > the name `super_good-solidus_taxjar` breaks some things like install generators installing migrations due to the hyphen in the name. the migrations ported over will have `.super_good-solidus_tax_jar` appended to it which gives a `ActiveRecord::IllegalMigrationNameError` from the hyphen.

Ahh ok, that makes sense!

Since the migrations aren't introduced until #32, Nick and I can tackle solving that problem in a different PR, and get this one through, as it adds value in itself.

AdnanTheExcellent (approved):
LGTM

jarednorman (approved):

PR #44, opened 2021-03-07
Provide an ActiveJob implementation for TaxJar reporting

aldesantis:
What is the goal of this PR?
---

This PR provides a basic implementation of TaxJar reporting, based on ActiveJob.

NOTE: This is still a draft. I need some infrastructure work on the extension to be completed first (see #43).

Todos
---

- [ ] Enqueue DeleteOrderJob when an order is cancelled

---

TBD.

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

Screenshots
---
N/A.

### DIFF
--- a/app/decorators/super_good/solidus_taxjar/spree/order_updater/fire_recalculated_event.rb
+++ b/app/decorators/super_good/solidus_taxjar/spree/order_updater/fire_recalculated_event.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module SuperGood
+  module SolidusTaxjar
+    module Spree::OrderUpdater
+      module FireRecalculatedEvent
+        def persist_totals
+          Spree::Event.fire 'order_recalculated', order: order
+          super
+        end
+
+        ::Spree::OrderUpdater.prepend(self) if Spree.solidus_gem_version < Gem::Version.new('2.11.0')
+      end
+    end
+  end
+end
--- a/config/initializers/subscribers.rb
+++ b/config/initializers/subscribers.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+if Spree.solidus_gem_version < Gem::Version.new('2.11.0')
+  require SuperGoodSolidusTaxjar::Engine.root.join('app/subscribers/super_good/solidus_taxjar/reporting_subscriber')
+
+  SuperGood::SolidusTaxjar::ReportingSubscriber.subscribe!
+end
--- a/app/jobs/super_good/solidus_taxjar/create_order_job.rb
+++ b/app/jobs/super_good/solidus_taxjar/create_order_job.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module SuperGood
+  module SolidusTaxjar
+    class CreateOrderJob < ApplicationJob
+      queue_as { SuperGood::SolidusTaxJar.job_queue }
+
+      def perform(order)
+        SuperGood::SolidusTaxJar::API.new.create_transaction_for(order)
+      end
+    end
+  end
+end
--- a/app/jobs/super_good/solidus_taxjar/create_refund_job.rb
+++ b/app/jobs/super_good/solidus_taxjar/create_refund_job.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module SuperGood
+  module SolidusTaxjar
+    class CreateRefundJob < ApplicationJob
+      queue_as { SuperGood::SolidusTaxJar.job_queue }
+
+      def perform(reimbursement)
+        SuperGood::SolidusTaxJar::API.new.create_refund_for(reimbursement)
+      end
+    end
+  end
+end
--- a/app/jobs/super_good/solidus_taxjar/delete_order_job.rb
+++ b/app/jobs/super_good/solidus_taxjar/delete_order_job.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module SuperGood
+  module SolidusTaxjar
+    class DeleteOrderJob < ApplicationJob
+      queue_as { SuperGood::SolidusTaxJar.job_queue }
+
+      def perform(order)
+        SuperGood::SolidusTaxJar::API.new.delete_transaction_for(order)
+      end
+    end
+  end
+end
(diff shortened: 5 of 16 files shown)


jarednorman:
@nvandoorn @Noah-Silvera Have you two taken a look at this?

Noah-Silvera:
> @nvandoorn @Noah-Silvera Have you two taken a look at this?

Just had a look! This seems like a good start.

I think we want to work towards hooking this reporting into the Orders/Refunds flow in solidus by default. The Taxjar certification checklist implies that users should be able to turn reporting "on and off", which we interpret as reporting should be baked into the extension, with configuration to enable or disable it.

aldesantis:
@Noah-Silvera @jarednorman this should be ready for another look! It now integrates directly with the Solidus event bus to enqueue the background jobs when orders are created/updated/reimbursed. I also added a feature toggle to easily enable/disable automatic reporting. The reporting hooks respect the `taxable_order_check` option, so that non-taxable orders aren't reported to TaxJar.

A couple of missing pieces here:

- I'm not 100% sure whether we should delete orders in TaxJar when they are cancelled in Solidus. What are your thoughts? 
- I'm also not sure whether we should persist the TaxJar order ID/sync time in Solidus, to track which orders have been persisted to TaxJar and which haven't. In some client apps where we don't do this, we ran into an annoying issue where old orders are recalculated for some reason and we attempt to update them in TaxJar, but they wee never created in the first place and the TaxJar API throws an error.
- Finally, I don't really like the compatibility layer that I wrote for firing the order_recalculated event on Solidus 2.10, where it's not provided out of the box. If other extensions did the same, we could potentially fire the event multiple times. I think this can only be solved by providing the compatibility layer as a separate extension that all other extensions depend on, but it's something that will obviously take a bit of work. In the meantime, we could just use a plain old decorator on Solidus 2.10 instead of relying on the event bus.

Let me know your thoughts on the above!

config/initializers/subscribers.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+if Spree.solidus_gem_version < Gem::Version.new('2.11.0')
```
Noah-Silvera, line 3: Am I correct in saying this check is here because this gem does not currently support solidus 2.11 ?

If that's the case, we should probably just make that explicit in the gemfile, and remove this check, as reporting is a required feature for this gem's certification. (Making that limitation explicit could be done outside of this PR)

app/decorators/super_good/solidus_taxjar/spree/order_updater/fire_recalculated_event.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module SuperGood
+  module SolidusTaxjar
+    module Spree::OrderUpdater
```
Noah-Silvera, line 5: Could we include a commit body that states that this override is here because of a bug in solidus 2.10 and below?

Noah-Silvera (commented):
> I'm not 100% sure whether we should delete orders in TaxJar when they are cancelled in Solidus. What are your thoughts?

[This section of the TaxJar certification documentation](https://developers.taxjar.com/integrations/sales-tax-reporting/#section-order-transactions) tells me that we should delete cancelled orders from TaxJar

"While you can import orders at any time, we recommend importing an order once it’s been shipped and paid/invoiced. If an order is canceled, you can avoid having to delete the order from TaxJar if you wait until it’s been paid."

> I'm also not sure whether we should persist the TaxJar order ID/sync time in Solidus, to track which orders have been persisted to TaxJar and which haven't. In some client apps where we don't do this, we ran into an annoying issue where old orders are recalculated for some reason and we attempt to update them in TaxJar, but they wee never created in the first place and the TaxJar API throws an error.

TaxJar has some guidelines around backfilling transactions if they were originally posted to TaxJar. They recommend the pattern of defaulting to a "Create order" request if the order is not found in the "Update order" request, which sounds like a reasonable way to deal with this problem that doesn't require us storing the TaxJar order ID/sync time in Solidus.

> Finally, I don't really like the compatibility layer that I wrote for firing the order_recalculated event on Solidus 2.10, where it's not provided out of the box. If other extensions did the same, we could potentially fire the event multiple times. I think this can only be solved by providing the compatibility layer as a separate extension that all other extensions depend on, but it's something that will obviously take a bit of work. In the meantime, we could just use a plain old decorator on Solidus 2.10 instead of relying on the event bus.

Updating the order multiple times should be idempotent, so I think this would be more of a problem in terms of spamming the job queue. I don't really like it either, but it does guarantee that the transaction is updated in TaxJar, which is the most important part. I think even if it was moved to a compatibility layer as a separate extension that other extensions depend on, we can't rely on all other extensions the consumer has installed all using the compatibility layer we create, so we still have the  potential problem of it firing multiple times. 

However, I think documenting that this extension does include this compatibility layer, and providing a way to disable it / instructions on how to disable it is important. Even though in our situation firing the event multiple times shouldn't cause problems, our code inadvertently firing the even multiple times could cause problems for consuming codebases that use that event, so we should provide a way to prevent our decorator from being loaded.

Noah-Silvera:
Hey @aldesantis ! I just wanted to check in on this PR. Getting this through would help push some other work towards certification of this extension forward, so we wanted to ask if you’d be up for @nvandoorn  and I jumping in on this PR and building on top of the work you’ve done.
 
Do you have any in progress work on this locally that we would be missing if we were to fork this?

aldesantis:
Hey @Noah-Silvera, absolutely, feel free to make it yours and continue the work here. I do apologize for disappearing, but the last few weeks have been crazy and I haven't really had a chance to finalize it...

app/decorators/super_good/solidus_taxjar/spree/order_updater/fire_recalculated_event.rb
aldesantis: It's not a bug, it's more a missing feature. 🙂 We only started firing the order_recalculated event in Solidus 2.11, so on Solidus < 2.11 we need to fire it ourselves.

config/initializers/subscribers.rb
aldesantis: See my comment below. We do support Solidus 2.11, this is just a "polyfill" for adding a missing feature (the order_recalculated event) which was only introduced in Solidus 2.11

Noah-Silvera:
> Hey @Noah-Silvera, absolutely, feel free to make it yours and continue the work here. I do apologize for disappearing, but the last few weeks have been crazy and I haven't really had a chance to finalize it...

@aldesantis thanks! Totally understand that, and we appreciate you the work you've done on this so far! We'll make sure to tag you on the PR we make out of forking this :)

PR #45, opened 2021-03-10
Update the engine name

Noah-Silvera:
What is the goal of this PR?
---

Once we add an install generator with migrations, having an engine name with a dash in it will cause the migration generation to fail, as migrations will be created with the engine name in it, and dashes are not allowed.

### DIFF
--- a/lib/super_good/engine.rb
+++ b/lib/super_good/engine.rb
@@ -3,6 +3,6 @@
 module SuperGoodSolidusTaxjar
   class Engine < Rails::Engine
     isolate_namespace Spree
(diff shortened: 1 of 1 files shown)


Noah-Silvera:
@jarednorman @AdnanTheExcellent this should remove the need to rename the gem once #32 adds an install generator.

adammathys (approved):

jarednorman (approved):

jarednorman:
Aha! Awesome, so we don't have to rename the gem!

Issue #28, opened 2020-11-05
The Spree::Address#address2 attribute is ignored during address validation


jasonwmatthews:
The `SuperGood::SolidusTaxJar::APIParams` module does not include the `Spree::Address#address2` attribute when constructing the `street` parameter for TaxJar's API. The `street` parameter is defined as the "[freeform input](https://developers.taxjar.com/api/reference/#post-validate-an-address)" indicating the two Solidus address fields should be concatenated if both are populated.

Omitting the `address2` attribute can result in slightly different normalized addresses where an apartment number, mail stop, or other unit information has a unique ZIP+4 code.

One fairly large example, One World Trade Center in New York has around 40 unique ZIP+4 codes, depending on floor or suite number:

1 WORLD TRADE CTR
NEW YORK NY 10007-0089

1 WORLD TRADE CTR STE 45A
NEW YORK NY 10007-0093

1 WORLD TRADE CTR FL 50
NEW YORK NY 10007-0101

1 WORLD TRADE CTR FL 99
NEW YORK NY 10007-0089

Noah-Silvera:
This was released in 0.18.2

PR #64, opened 2021-04-30
Use spree address2 if it is present

Noah-Silvera:
What is the goal of this PR?
---

The taxjar validation API has the option to take a freeform street input consisting of the full address. This means that if both address1 and address2 exist, we should concatenate them and pass them to the street parameter.

This can help with situations that involve a ZIP+4 code in the US for example. A user may enter their 5 digit zip code in the zip code field, but enter their ZIP+4 code in the address2 field.

---

1. Manually call this API with a valid taxjar key and the situation described above with a ZIP+4 code. Ensure the address is validated correctly with the ZIP+4 code in the freeform input taken into account

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#59](https://github.com/SuperGoodSoft/solidus_taxjar/pull/59) Add pry debugging tools
 - [#69](https://github.com/SuperGoodSoft/solidus_taxjar/pull/69) Lock ExecJS version
 - [#37](https://github.com/SuperGoodSoft/solidus_taxjar/pull/37) Added a basic Taxjar settings admin interface which displays placeholder text.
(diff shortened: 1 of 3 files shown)


Noah-Silvera:
This closes #28 
cc: @jasonwmatthews

adammathys (approved):

jarednorman (commented):
Needs changelog update, but otherwise looks good.

Noah-Silvera:
> Needs changelog update, but otherwise looks good.

Updated the changelog!

jarednorman (approved):

bl-ue:
>but otherwise looks good

@jarednorman you mean "both otherwise looks super good"? 😆

Noah-Silvera:
This is completed and released in 0.18.2

PR #85, opened 2021-06-23
Fix invalid refund endpoint call

Noah-Silvera:
What is the goal of this PR?
---

The refund endpoint is meant to be used to post a transaction that represents the delta of changes to the order. This means that the amounts for the order total and sales tax passed should be negative.

An example can be found in the taxjar API

https://developers.taxjar.com/api/reference/?shell#post-create-a-refund-transaction

---

1. Manually trigger the `create_refund_for` endpoint on the extension 
    * [ ] Ensure a refund with the appropriate negative values is created on your taxjar account

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

Screenshots
---


Noah-Silvera:
Before updating the changelog, we should release the current changes we have on TaxJar, as they are currently non-breaking.

This is a breaking change we should do a separate version release for.

Noah-Silvera:
Closing in favour of #86

PR #86, opened 2021-06-23
Fix invalid refund endpoint call

Noah-Silvera:
What is the goal of this PR?
---

The refund endpoint is meant to be used to post a transaction that represents the delta of changes to the order. This means that the amounts for the order total and sales tax passed should be negative.

An example can be found in the taxjar API

https://developers.taxjar.com/api/reference/?shell#post-create-a-refund-transaction

---

1. Manually trigger the `create_refund_for` endpoint on the extension 
    * [x] Ensure a refund with the appropriate negative values is created on your taxjar account

Merge Checklist
---

- [x] Run the manual tests
- [ ] Update the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master
 
 - [#88](https://github.com/SuperGoodSoft/solidus_taxjar/pull/88) Fire `shipment_shipped` event when any shipment on an order ships.
+- [#86](https://github.com/SuperGoodSoft/solidus_taxjar/pull/86)  Fix invalid refund endpoint call.
 
 ## v0.18.2
 
--- a/lib/super_good/solidus_taxjar/api_params.rb
+++ b/lib/super_good/solidus_taxjar/api_params.rb
@@ -60,9 +60,9 @@ module SuperGood
               transaction_id: reimbursement.number,
               transaction_reference_id: reimbursement.order.number,
               transaction_date: reimbursement.order.completed_at.to_formatted_s(:iso8601),
-              amount: reimbursement.total - additional_taxes,
+              amount: -1 * (reimbursement.total - additional_taxes),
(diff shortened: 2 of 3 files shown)


Noah-Silvera:
Before updating the changelog, we should release the current changes we have on TaxJar, as they are currently non-breaking.

This is a breaking change we should do a separate version release for.

adammathys (approved):

forkata (approved):

Noah-Silvera:
I found a conflicting KB article posted in 2021 that says we should use positive values positing a refund :(

Will need to reach out to the taxjar team to check which pattern is actually the correct pattern

https://support.taxjar.com/article/351-handling-refunds-with-api

nvandoorn:
Thanks @Noah-Silvera! Should I start a thread with the TaxJar folks over email?

Noah-Silvera:
> Thanks @Noah-Silvera! Should I start a thread with the TaxJar folks over email?

It's probably worth it just to be sure! I found one more piece of information that strengthens the case for negative values

In the taxjar API docs for create order, it says this

Screen Shot 2021-06-25 at 15 20 45PDT

And in the taxjar API docs for create refund, it says this

Screen Shot 2021-06-25 at 15 20 38PDT


I am 80% confident the KB article I linked is wrong, and would be comfortable merging this for now, but waiting to release this change until we have gotten 100% confidence from them that we should use negative values.

jarednorman (approved):

jarednorman (approved):

forkata (approved):

nvandoorn (approved):

Noah-Silvera:
We are deprecating this refund endpoint in favour of one that refunds based on a taxjar transaction and not a reimbursement, so we can close this PR.

Issue #77, opened 2021-05-21
Support reporting


nvandoorn:
At this moment, this extension only calculates tax and does not report anything to TaxJar. This is a large feature that will require further decomposition.

Noah-Silvera:
After further discussion and research, we've landed on some acceptance requirements for how this will work.

[According to the TaxJar documentation](https://developers.taxjar.com/integrations/sales-tax-reporting/), we only need to push orders in certain cases.

> While you can import orders at any time, we recommend importing an order once it’s been shipped and paid/invoiced. If an order is canceled, you can avoid having to delete the order from TaxJar if you wait until it’s been paid. There’s a couple ways you may want to do this:
> If your platform has a completed_at date, only import orders that have been completed.
> If your platform has the notion of an order status, ensure the order status is in a shipped, paid, or complete status before importing.

We should only push orders if all the following are true...
     - `SuperGood::SolidusTaxJar.taxable_order_check.call(order)` 
     - `order.complete?
     - `order.paid?

     * As long as an order has **not** been paid, we don't need to worry about pushing it to TaxJar, as no actual monetary transaction has been made yet sent - the client has not been charged yet. (We may want to confirm this with the TaxJar team - just to confirm it's ok that unpaid and unshipped orders on the system never get pushed to TaxJar).
     
     
In order for an order to be paid and shipped, "order_recalculated" has to be fired at least once. This means we only need to hook into refunds being created, and orders being recalcalculated.

Remaining questions: 

1. What's the different between refunds and reimbursements? Which ones do we care about? How can we hook into solidus to report to TaxJar when they happen? 
2. What about store credit refunds?

Noah-Silvera:
https://support.taxjar.com/article/250-how-are-refunds-displayed-in-my-reports

Noah-Silvera:
Investigated with @forkata and learned a little bit more about how refunds work, and had some more thoughts on how we need to handle them to correctly report them to TaxJar.

Some Background
-----

We learned in #86 that refunds are actually a negative delta of the amounts refunded from the original order. An order will have at most two transactions associated with it - an order transaction, which represents all the positive payments made towards the order towards what items, and a refund transaction that represents the store-credit/credit card refunds given back to a customer.

Assumptions
-----

* TaxJar only needs an order/refund to be reported when a financial transaction is actually completed.  
[...]

PR #88, opened 2021-07-19
#77 Add shipment shipped event

Noah-Silvera:
What is the goal of this PR?
---

An an event for when any shipment ships. Orders need to be pushed to TaxJar whenever any part of the order is shipped, so we hook into solidus to add a new event to support this. 

Merge Checklist
---

- [x] Update the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## master
+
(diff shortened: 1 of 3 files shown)


adammathys (approved):

jarednorman (approved):

forkata (approved):
Thanks Noah, this would make a great PR to Solidus!

benjaminwil (approved):

nvandoorn (approved):

PR #95, opened 2021-08-10
Only require state for Canada and US addresses

alexblackie:
What is the goal of this PR?
---

According to the TaxJar API, the `to_state` field is only required for
CA or US addresses. Right now, we consider an address incomplete if it
does not have a state, which breaks calculation of EU taxes as it is
commonplace to not have a "state"-equivalent.

We should change the incomplete check to only require the state if
TaxJar requires it for that country. This will allow calculation of tax
for EU addresses.

---

Use an EU address with no "state", ensure taxes are still calculated.

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog


jarednorman (approved):

forkata (approved):
❤️

Noah-Silvera (commented):
For the record, I did review this before merging it 😂

nvandoorn (commented):
testing

Issue #89, opened 2021-07-21
Add support for fetching a transaction through API


forkata:
## What

Implement `show_order` transaction API - 
https://developers.taxjar.com/api/reference/#get-show-an-order-transaction.

## Why

We need to check if a transaction already exists on TaxJar before attempting to create it. This will allow us to attempt to fetch it by the `order_number`.

PR #97, opened 2021-08-18
Add taxjar show order endpoint

Noah-Silvera:
What is the goal of this PR?
---

Add an endpoint to show an order #89 

Manual Tests
---

* [x] Manually call this endpoint and ensure it returns an expected response.

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

### DIFF
--- a/lib/super_good/solidus_taxjar/api.rb
+++ b/lib/super_good/solidus_taxjar/api.rb
@@ -50,6 +50,10 @@ module SuperGood
         taxjar_client.delete_order order.number
       end
 
+      def show_latest_transaction_for(order)
+        taxjar_client.show_order order.number
+      end
+
       def create_refund_for(reimbursement)
         taxjar_client.create_refund ApiParams.refund_params(reimbursement)
       end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#95](https://github.com/SuperGoodSoft/solidus_taxjar/pull/95) Only require "state" for Canadian and US addresses
 - [#98](https://github.com/SuperGoodSoft/solidus_taxjar/pull/98) Add generator for TaxJar transaction IDs
 - [#103](https://github.com/SuperGoodSoft/solidus_taxjar/pull/103) Add `reporting_enabled` configuration option
+- [#97](https://github.com/SuperGoodSoft/solidus_taxjar/pull/97) Add public API method to request the latest transaction associated with a solidus order.
 
 ## v0.18.2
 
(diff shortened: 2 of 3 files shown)


jarednorman (approved):

lib/super_good/solidus_taxjar/api.rb
```diff
@@ -47,6 +47,10 @@ def delete_transaction_for(order)
         taxjar_client.delete_order order.number
       end
 
+      def show_transaction_for(order)
```
forkata, line 50: I am curious if the API here should be to pass the `Spree::Order` record or take the `transaction_id` directly. The reason I am asking that is that only in the case of the initial transaction can we infer the `id` from the order number, for subsequent transactions created for the order (after an update) we won't have the necessary context and may need to just rely on the `transaction_id` to be provided 🤔 

For extra context, me and @benjaminwil  are working on a class to help us generate initial and subsequent transaction id's and we realized we need to have the context of what the `transaction_id` is on TaxJar for an order in order to do that.

forkata (commented):
Thanks @Noah-Silvera and @nvandoorn! This change looks good, I have one question here about the parameter to this call.

lib/super_good/solidus_taxjar/api.rb
benjaminwil: If we don't store `transaction_id`s in the DB I can't think of a way around this.

We might have to fetch this transaction, then find out if it's been cancelled and fetch the most recent transaction that contains the order number?

forkata: > We might have to fetch this transaction, then find out if it's been cancelled and fetch the most recent transaction that contains the order number?

I am not sure we can query TaxJar in that way, looking at the documentation for the `show_order` API, it takes a `transaction_id` and gives you back the transaction corresponding to that. So we may have to start with the initial one and keep fetching until we find one that isn't cancelled, but that's exactly what I am aluding to here, we'll inevitably have to fetch a transaction by it `transaction_id` which may be different from the `order.number`.

forkata: I am okay with that! We may have to change the return type because there could be multiple transactions for a single order, but we can deal with that at a later point. I am fine keeping this consistent with the other methods and handling the actual fetching by the `transaction_id` in the implementation.

forkata (approved):

Noah-Silvera:
This last push just updated the method name to be more clear about which transaction it's getting for an order (the latest one!)

benjaminwil (approved):

jarednorman (approved):

Noah-Silvera:
just a changelog rebase!

Issue #94, opened 2021-07-29
TaxJar `transaction_id` generator


forkata:
## Why
We need a utility to help us generate `transaction_id` identifiers for TaxJar based on the Solidus order number for the case where a past transaction has been refunded and we want to re-create a new transaction with the current state of the order in Solidus. 

TaxJar does not provide a way for us to associate transactions (other than for refunds) so we want to use the `transaction_id` field to create that association.

## How

Implement a utility class which can be used to generate the `transaction_id` from an order based on the following criteria. This should work for both the initial transaction as well as any re-created transactions after a Solidus order recalculation.

* if a `previous_transaction_id` is not provided
   * return the `solidus_order.number`
* if a `previous_transaction_id` is provided
   * and it does not have a suffix
     * return `solidus_order.number-1`
   * and it has a suffix, i.e. `R01234567890-1`
     * increment the suffix and return `solidus_order.number-#{new suffix}`

PR #98, opened 2021-08-20
TaxJar Reporting | Transaction ID generator

forkata:
What is the goal of this PR?
---
Finishes #94 

Introduce a utility class for generating `transaction_id`s for transactions
created using the TaxJar reporting API.

This is a step towards implementing a way to update transactions on TaxJar after
they have been updated in Solidus. The recommended way to reflect changes is to
cancel the original transaction and create a new one while referencing the old
one. However TaxJar does not have a way to associate two transactions in any way
so we will be using a suffix appended to the original order number and increment
that every time we cancel and re-create a transaction because the Solidus order
changed. This utility will aid in genrating a new `transaction_id` based on the
current one in TaxJar.

---

1. Create a transaction on TaxJar using the API and confirm the `transaction_id`
  is the same as the `order.number`.
1. For now the method to generate a new `transaction_id` from a previous one is
  not used anywhere so testing that manually requires you to call the class on
  the console.

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

### DIFF
--- a/lib/super_good/solidus_taxjar/api_params.rb
+++ b/lib/super_good/solidus_taxjar/api_params.rb
@@ -37,13 +37,13 @@ module SuperGood
           }.merge(order_address_params(address))
         end
 
-        def transaction_params(order)
+        def transaction_params(order, transaction_id = order.number)
           {}
             .merge(customer_params(order))
             .merge(order_address_params(order.tax_address))
             .merge(transaction_line_items_params(order.line_items))
             .merge(
-              transaction_id: order.number,
+              transaction_id: transaction_id,
               transaction_date: order.completed_at.to_formatted_s(:iso8601),
               amount: [order.total - order.additional_tax_total, 0].max,
               shipping: shipping(order),
--- a/lib/super_good/solidus_taxjar/transaction_id_generator.rb
+++ b/lib/super_good/solidus_taxjar/transaction_id_generator.rb
@@ -1,0 +1,45 @@
+module SuperGood
+  module SolidusTaxjar
+    # Responsible for generating `transaction_id` references for transactions
+    # we create on TaxJar for Solidus orders. This class handles creating
+    # associated ID's for transactions which need to be cancelled and recreated
+    # when the order is updated in Solidus after it has been sent to the TaxJar
+    # reporting API.
+    class TransactionIdGenerator
+      class << self
+        # Generates the next sequential `transaction_id` given an order and
+        # optionally the current transaction ID on TaxJar. This handles the
+        # case where a transaction already has been created on TaxJar and later
+        # needs to be cancelled and we need to create an updated transaction
+        # with an associated identifier.
+        #
+        # @param order [Spree::Order] the order for which we want to generate a
+        #   transaction ID.
+        # @param current_transaction_id [String] the current transaction ID for
+        #   the order if it exists on TaxJar.
+        # @return [String] the next sequential `transaction_id`
+        def next_transaction_id(order:, current_transaction_id: nil)
+          if current_transaction_id.nil?
+            "#{order.number}"
+          elsif order.number == current_transaction_id
+            "#{current_transaction_id}-1"
+          else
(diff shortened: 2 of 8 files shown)


lib/super_good/solidus_taxjar/transaction_id_generator.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        # Generates the next sequential `transaction_id` from the current one
+        # for the transaction on TaxJar. We do this if the current transaction
+        # has been cancelled and we need to create an updated transaction.
+        #
+        # @param current_transaction_id [String]
+        # @return [String] the next `transaction_id`
+        def next_transaction_id(current_transaction_id)
+          parts = current_transaction_id.split(/-/)
+          if parts.length == 1
```
Noah-Silvera, line 25: If the order number has a dash in it, this falls apart.

Could we do if `current_transaction_id == order.number` instead to deal with the base case?

Noah-Silvera (approved):

Noah-Silvera (changes requested):
forgot to request changes

lib/super_good/solidus_taxjar/transaction_id_generator.rb
forkata: Great catch, we considered that for the second case but not for this initial one. I like your recommendation, we'll add a spec for this and update the change to handle that case!

Noah-Silvera (approved):

lib/super_good/solidus_taxjar/api_params.rb
```diff
@@ -37,13 +37,13 @@ def tax_rate_address_params(address)
...
 
-        def transaction_params(order)
+        def transaction_params(order, transaction_id = order.number)
           {}
             .merge(customer_params(order))
             .merge(order_address_params(order.tax_address))
             .merge(transaction_line_items_params(order.line_items))
             .merge(
-              transaction_id: order.number,
+              transaction_id: transaction_id,
```
Noah-Silvera, line 46: Added a comment to #101 around updating this to be the orders latest/current transaction ID once the work to associate an order with taxjar transactions has been completed.

nvandoorn (approved):

lib/super_good/solidus_taxjar/transaction_id_generator.rb
```diff
@@ -0,0 +1,45 @@
...
+        # @param current_transaction_id [String] the current transaction ID for
+        #   the order if it exists on TaxJar.
+        # @return [String] the next sequential `transaction_id`
+        def next_transaction_id(order:, current_transaction_id: nil)
+          if current_transaction_id.nil?
+            "#{order.number}"
+          elsif order.number == current_transaction_id
+            "#{current_transaction_id}-1"
+          else
+            parts = current_transaction_id.split(/-/)
```
adammathys, line 27: You could use [`rpartition`](https://ruby-doc.org/core-2.5.1/String.html#method-i-rpartition) here instead which would remove the need for the `[0..-2].join("-")` below.

lib/super_good/solidus_taxjar/transaction_id_generator.rb
```diff
@@ -0,0 +1,45 @@
...
+            enumerator = parts.last.to_i
+            "#{parts[0..-2].join("-")}-#{enumerator + 1}"
```
adammathys, line 29: Ruby's String class has a [`next`](https://ruby-doc.org/core-2.5.1/String.html#method-i-next) method that does exactly this for us, which removes the need for converting to/from an integer.

adammathys: Which combined with the above comment could lead to this simplification of this code:

```ruby
parts = current_transaction_id.rpartition('-')
parts.last.next!
parts.join
```

nvandoorn (approved):

Noah-Silvera (approved):

PR #104, opened 2021-09-01
Update readme local development instructions

Noah-Silvera:
What is the goal of this PR?
---

The local development instructions were unclear and missing required environment variables, and how to use / the existence of the `solidus-dev-support` tooling is not obvious. This PR intends to correct and ease the setup of your local environment variables, and add a blurb on testing and the sandbox app and the gotchas related to `solidus-dev-support`.    

Merge Checklist
---

- [x] Update the changelog

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,0 +1,6 @@
+export SOLIDUS_BRANCH=v2.10
+export DB=postgresql
+export RAILS_VERSION="~> 5.2.0"
+export TAXJAR_API_KEY="your-api-key-here"
+# Uncomment the following line to run the app against the sandbox API
+# export TAXJAR_API_URL="https://api.sandbox.taxjar.com"
--- a/README.md
+++ b/README.md
@@ -89,22 +89,35 @@ endpoint of the tax calculator in order to provide the most coherent and
 reliable results. TaxJar support recommends using this endpoint for live
 calculations.
 
-## Development
+## Development Setup
 
-Before installing dependencies, make sure your environment variables are setup.
-Our CI uses something like this:
-`SOLIDUS_BRANCH=v2.10 DB=postgresql RAILS_VERSION="~> 5.2.0"`.
+The environment variables needed for development are listed in `.env.sample`. Copy these to `.env`, replace the appropriate values, and run `source .env` in your terminal.
+  * **NOTE: These environment variables need to be set when installing dependencies, running specs, and running the sandbox app**
+  * You can get a TaxJar API key by [signing up for a free trial](https://app.taxjar.com/sign_up).
 
-**NOTE: These environment variables need to be set when installing dependencies
-and running specs, so set them before each command like this:**
+After setting the environment variables, run `bin/setup` to install dependencies. These dependencies should match the rails/solidus versions you specified in your environment variables.
 
-```sh
-SOLIDUS_BRANCH=v2.10 DB=postgresql RAILS_VERSION="~> 5.2.0" bin/setup
-```
 
-Run `bin/setup` to install dependencies. Then, run `bundle exec rake` to run the
-tests. You can also run `bin/console` for an interactive prompt that will allow
-you to experiment.
+### Running the Tests
+
+The tests are written in rspec, and can be run with `bundle exec rspec`
+
+* If this is the first time you've run the tests, you will notice that a "dummy" rails application is created in `spec/dummy`
+* **NOTE: If you change the `RAILS_VERSION` or `SOLIDUS_BRANCH` environment variables, you will need to delete this folder before running the tests again**
+
+### Running a local Solidus app with TaxJar
+
+This extension uses the `solidus_dev_support` gem to provide tooling for developing extensions. One of it's most important features is the sandbox app, which runs a solidus store with this extension hooked in.
+
+https://github.com/solidusio/solidus_dev_support#sandbox-app
+
+By using `bin/rails-sandbox`, you can run any command that `bin/rails` would accept (`bin/rails-sandbox server`, `bin/rails-sandbox console`), except it will run it on a locally generated sandbox store.
+
+* `binding.pry`s set in your local source code will be hit by the sandbox app, as it links itself directly to your development folder.
+* If this is the first time you've run the sandbox, you will notice that a `sandbox` folder is created with a solidus store.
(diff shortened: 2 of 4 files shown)


Noah-Silvera:
@senemsoy would love for you to have a look at this and tell me if it's more clear!

README.md
```diff
@@ -89,22 +89,35 @@ endpoint of the tax calculator in order to provide the most coherent and
...
+
+https://github.com/solidusio/solidus_dev_support#sandbox-app
+
+By using `bin/rails-sandbox`, you can run any command that `bin/rails` would accept (`bin/rails-sandbox server`, `bin/rails-sandbox console`), except it will run it on a locally generated sandbox store.
+
+* `binding.pry`s set in your local source code will be hit by the sandbox app, as it links itself directly to your development folder.
+* If this is the first time you've run the sandbox, you will notice that a `sandbox` folder is created with a solidus store.
+* **NOTE: If you change the `RAILS_VERSION` or `SOLIDUS_BRANCH` environment variables, you will need to delete this folder before running the tests again**
+
+### Releasing a new version
```
Noah-Silvera, line 120: @jarednorman I moved this into its own section, but it brought up the question... are these instructions accurate? So far to my memory...

1. We update the changelog to display the new version instead of `master`
2. You run the script to release it.
    * Is this a secret script? Should we remove these instructions altogether and just keep that knowledge/keys to release limited to you and maybe another person or two?

jarednorman: The instructions in the README are not accurate, but your instructions in your comment basically are.

```sh
gem bump -v [major|minor|patch] -p -t -r
```

jarednorman: Using this thing: https://github.com/svenfuchs/gem-release

README.md
```diff
@@ -89,22 +89,35 @@ endpoint of the tax calculator in order to provide the most coherent and
 reliable results. TaxJar support recommends using this endpoint for live
 calculations.
 
-## Development
+## Development Setup
```
benjaminwil, line 92: Chris recently linted the README to have hard-wrapped lines at 80 characters. Though it is technically inconsequential, my preference would be to continue doing that.

README.md
```diff
@@ -89,22 +89,35 @@ endpoint of the tax calculator in order to provide the most coherent and
...
+### Running the Tests
+
+The tests are written in rspec, and can be run with `bundle exec rspec`
+
+* If this is the first time you've run the tests, you will notice that a "dummy" rails application is created in `spec/dummy`
+* **NOTE: If you change the `RAILS_VERSION` or `SOLIDUS_BRANCH` environment variables, you will need to delete this folder before running the tests again**
+
+### Running a local Solidus app with TaxJar
+
+This extension uses the `solidus_dev_support` gem to provide tooling for developing extensions. One of it's most important features is the sandbox app, which runs a solidus store with this extension hooked in.
```
benjaminwil, line 110:
```suggestion
This extension uses the `solidus_dev_support` gem to provide tooling for developing extensions. One of its most important features is the sandbox app, which runs a solidus store with this extension hooked in.
```

Sorry. 😂

senemsoy (approved):
Looks great to me!

README.md
```diff
@@ -89,22 +89,35 @@ endpoint of the tax calculator in order to provide the most coherent and
 reliable results. TaxJar support recommends using this endpoint for live
 calculations.
 
-## Development
+## Development Setup
 
-Before installing dependencies, make sure your environment variables are setup.
-Our CI uses something like this:
-`SOLIDUS_BRANCH=v2.10 DB=postgresql RAILS_VERSION="~> 5.2.0"`.
+The environment variables needed for development are listed in `.env.sample`. Copy these to `.env`, replace the appropriate values, and run `source .env` in your terminal.
```
forkata, line 94: Curious how others feel about this as a step we encourage for development on the extension. It is quite verbose to specify the full environment before each command, but at least it is explicit. I am worried this may introduce issues for people who are not used to having to source this file every time before running a command in a new shell since we don't use `dotenv` in this gem.

benjaminwil: Yeah, I might be more inclined to just provide a list of all of the environment variable combinations that are under CI, and then let users copy-paste those as a convenience.

I will personally just be including the environment variables in my commands.

We could always provide `dotenv` as a development gem, too.

jarednorman: People should just use direnv. 😅

jarednorman (approved):

README.md
forkata: I liked one suggestion @jarednorman had which is to update the defaults / fallback for these environment variables to things we consider "sane" in the `Gemfile`. The defaults we have here don't seem appropriate for the current state of the gem
https://github.com/SuperGoodSoft/solidus_taxjar/blob/master/Gemfile#L5
https://github.com/SuperGoodSoft/solidus_taxjar/blob/master/Gemfile#L12

.env.sample
```diff
@@ -0,0 +1,6 @@
+export SOLIDUS_BRANCH=v2.10
+export DB=postgresql
+export RAILS_VERSION="~> 5.2.0"
```
jarednorman, line 3: There are already defaults in the Gemfile. If you want to update them you should do that there.

README.md
jarednorman: These instructions won't work in all shells. 🐟 

I think rather than telling people "run this command", it's fine to say set these variables (since there's only one required and one optional.) You can provide the commands to export them in the README, but using a `.env` file is confusing:

- It might lead anyone familiar with dotenv to think that file would actually be loaded by something, which isn't the case.
- The file doesn't match the format of `.env` files, but of `direnv` files, which would be named `.envrc`.

Why don't we just tell people to set the two env variables and offer an example command to do so, in case the person doesn't know how

-----

@forkata You might consider using "sensible" instead of "sane". I'd hardly consider the existing defaults "insane". 😅

jarednorman (changes requested):

README.md
forkata: @jarednorman sorry for the bad paraphrase. As usual your choice of words is better than mine. I share the same concerns around trying to introduce a dotfile which may imply certain convention or use of a specific gem to manage the environment. If we wanted to provide a shell script for convenience that can be probably made to work independent of the shell someone is using, but even that is a bit too opinionated for me.

I think we should guide developers to what environment they need to export or specify before each command and allow them to set that in any way they wish. I also think this would not be as big of an issue if the defaults we provided actually produced a working gem, which is not the case now.

Noah-Silvera:
Some other of the changes have been moved into #121

Issue #101, opened 2021-08-25
Associate a spree_order with a taxjar transaction ID


Noah-Silvera:
An order in solidus can potentially have multiple transaction ID if it is refunded partially.

These transaction ID's will have sequential numbering, i.e. `order-number-transaction-1`, `order-number-transaction-2`, `order-number-transaction-3`. 

We want to ...

1. Associate these transaction ID's with the order, and
2. Be able to retrieve the latest transaction ID associated with the order

Noah-Silvera:
We should also update the default `transaction_id` used in the `transaction_params` to be the current associated transaction id once this work is completed!

Noah-Silvera:
We should decorate the spree order with a `latest_created_taxjar_transaction` method to return the latest `taxjar_transaction` that was created (not refunds) and use this in the various locations we currently use the `order.number` as the transaction_id.

forkata:
@Noah-Silvera Is there an API we can create which doesn't require us to decorate the `Spree::Order` model. I feel like adding more methods to `Order` is rarely the right answer? Maybe we can introduce a `Reporting` specific class which can give us all transactions, just the last one, refund transactions for a given order.

Noah-Silvera:
> @Noah-Silvera Is there an API we can create which doesn't require us to decorate the `Spree::Order` model. I feel like adding more methods to `Order` is rarely the right answer. Maybe we can introduce a `Reporting` specific class which can give us all transactions, just the last one, refund transactions, etc. for a given order. I feel like we may need more functionality in the future and starting down this path would probably mean more methods we decorate the `Order` with.

You solved this problem nicely Chris with adding a method to the `TaxjarTransaction` model!

https://github.com/SuperGoodSoft/solidus_taxjar/blob/2b12eaef399f4fc681db9da9588b40dcb6567e1c/app/models/super_good/solidus_taxjar/order_transaction.rb#L9

Noah-Silvera:
> We should also update the default transaction_id used in the transaction_params to be the current associated transaction id once this work is completed!

`ApiParams.transaction_params` is only used in two places. One of those places (`create_transaction_for`) has been updated to use the latest transaction id, which is appropriate. The other place (`update_transaction_for`) is an API call we don't currently plan to use in our default reporting implementation, and shouldn't change unnecessarily at this point, as it is a public API others may be relying on for their own reporting implementation.

PR #108, opened 2021-09-09
Add Taxjar transaction Model

Noah-Silvera:
What is the goal of this PR?
---

Orders can be associated with multiple taxjar transactions, so we should add a new model to keep track of this relationship. 

---

1. Create a new `SuperGood::SolidusTaxjar::OrderTransaction` model
   * [x] Ensure it validates the presence of an order and `transaction_id`
   * [x] Ensure calling `taxjar_order_transactions` on the `Spree::Order` returns an association with the new transaction

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

This ticket is the first step to completing #101.

### DIFF
--- a/app/models/super_good/solidus_taxjar/order_transaction.rb
+++ b/app/models/super_good/solidus_taxjar/order_transaction.rb
@@ -1,0 +1,9 @@
+module SuperGood
+  module SolidusTaxjar
+    class OrderTransaction < ActiveRecord::Base
+      belongs_to :order, class_name: "Spree::Order"
+
+      validates_presence_of :transaction_id
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - [#100](https://github.com/SuperGoodSoft/solidus_taxjar/pull/100) Add public API method to post a taxjar refund transaction for a solidus order.
 - [#102](https://github.com/SuperGoodSoft/solidus_taxjar/pull/102) Add description to transaction line item params
 - [#107](https://github.com/SuperGoodSoft/solidus_taxjar/pull/107) Fix rails-engine binstub to point to correct engine entry point
+- [#108](https://github.com/SuperGoodSoft/solidus_taxjar/pull/108) Add new model associated with a `Spree::Order` to represent taxjar order creation transactions
 
 ## v0.18.2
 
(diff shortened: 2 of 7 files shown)


Noah-Silvera (commented):
~~We should rename the table to `solidus_taxjar_order_transactions` as we will only be recording orders in it.~~

Done

benjaminwil:
Just noticed that one of the commits has `ben@super.gd` in the co-authors, but I'm `benjamin@super.gd`.

app/models/super_good/solidus_taxjar/transaction.rb
```diff
@@ -0,0 +1,7 @@
+module SuperGood
+  module SolidusTaxjar
+    class OrderTransaction < ActiveRecord::Base
+      belongs_to :order, class_name: "Spree::Order"
+    end
+  end
+end
```
benjaminwil, line 7: We need to move this file to `order_transaction.rb` in order to appease Zeitwerk's autoloading behaviour.

benjaminwil (commented):
One small change required.

app/models/super_good/solidus_taxjar/transaction.rb
Noah-Silvera: Fixed! Good Catch

Noah-Silvera:
> Just noticed that all of the commits have `ben@super.gd` in the co-authors, but I'm `benjamin@super.gd`.

Thank you for pointing this out! Updated.

app/models/super_good/solidus_taxjar/order_transaction.rb
```diff
@@ -0,0 +1,7 @@
+module SuperGood
+  module SolidusTaxjar
+    class OrderTransaction < ActiveRecord::Base
+      belongs_to :order, class_name: "Spree::Order"
```
forkata, line 4: Me and @benjaminwil did some testing and we think it would be a good idea to add a `validates_presence_of :transaction_id` here in addition to the DB-level validation we have.

forkata (commented):
Me and @benjaminwil pulled this down and ran through the manual testing instructions in the sandbox app. The only minor change we would make is add a Rails validation for the presence of the `transaction_id` otherwise this looks good to us 🚀

app/models/super_good/solidus_taxjar/order_transaction.rb
Noah-Silvera: Done :)

nvandoorn (approved):

benjaminwil (approved):

Noah-Silvera:
Last pushes just fixed a changelog conflict

PR #109, opened 2021-09-10
Save `OrderTransaction` after API call to TaxJar

forkata:
What is the goal of this PR?
---
This change creates a new record with the `transaction_id` reference used to identify the Solidus order on TaxJar when we create a transaction for it through the reporting API. This is necessary in order for us to be able to fetch and update the transaction in the future, especially in the case when there are multiple transactions for the same Solidus order (due to refund and updates) where the `order.number` does not correspond to the `transaction_id` on TaxJar.

---

1. Create a transaction for a Solidus order using the `Api.create_transaction_for(order)` method
    * [x] Check that `order.taxjar_order_transactions` has a record containing the `transaction_id` and a `transaction_date`.
    * [x] Fetch the TaxJar order transaction using the  `Api.show_latest_transaction_for(order)` and ensure this returns the correct TaxJar order transaction.

Merge Checklist
---
- [x] Use the persisted `transaction_id` for refund calls
- [x] Run the manual tests
- [x] Update the Changelog

Relates to #101

### DIFF
--- a/app/models/super_good/solidus_taxjar/order_transaction.rb
+++ b/app/models/super_good/solidus_taxjar/order_transaction.rb
@@ -5,6 +5,10 @@ module SuperGood
 
       validates_presence_of :transaction_id
       validates_presence_of :transaction_date
+
+      def self.latest_for(order)
+        where(order: order).order(transaction_date: :desc).limit(1).first
+      end
     end
   end
 end
--- a/lib/super_good/solidus_taxjar/api.rb
+++ b/lib/super_good/solidus_taxjar/api.rb
@@ -37,9 +37,14 @@ module SuperGood
 
       def create_transaction_for(order)
         transaction_id = TransactionIdGenerator.next_transaction_id(order: order)
-        taxjar_client.create_order(
+        response = taxjar_client.create_order(
           ApiParams.transaction_params(order, transaction_id)
         )
+        order.taxjar_order_transactions.create!(
+          transaction_id: response.transaction_id,
+          transaction_date: response.transaction_date
+        )
+        response
       end
 
       def update_transaction_for(order)
--- a/lib/super_good/solidus_taxjar/api.rb
+++ b/lib/super_good/solidus_taxjar/api.rb
@@ -56,7 +56,14 @@ module SuperGood
       end
 
       def show_latest_transaction_for(order)
-        taxjar_client.show_order order.number
+        latest_transaction_id =
+          OrderTransaction.latest_for(order)&.transaction_id
+
+        if latest_transaction_id.nil?
+          raise StandardError, "No latest TaxJar order transaction for #{order.number}"
+        end
+
+        taxjar_client.show_order(latest_transaction_id)
       end
 
       def create_refund_transaction_for(order)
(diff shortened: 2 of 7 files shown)


benjaminwil:
Right now, this PR is based off of #108

benjaminwil:
Rebased now that #108 is merged.

app/models/super_good/solidus_taxjar/order_transaction.rb
```diff
@@ -4,6 +4,10 @@ class OrderTransaction < ActiveRecord::Base
       belongs_to :order, class_name: "Spree::Order"
 
       validates_presence_of :transaction_id
+
+      def self.latest_for(order)
+        where(order: order).order(created_at: :desc).limit(1).first
```
Noah-Silvera, line 9: We are later going to need to backfill transactions... I could see this being incompatble with this method implementation, as we will create very old transactions in the backfill
 
https://developers.taxjar.com/integrations/sales-tax-reporting/

I think we might wanna record the `transaction_date` on our order_transaction model and sort by that instead.

lib/super_good/solidus_taxjar/api.rb
```diff
@@ -53,7 +53,9 @@ def delete_transaction_for(order)
       end
 
       def show_latest_transaction_for(order)
-        taxjar_client.show_order order.number
+        latest_transaction_id =
+          OrderTransaction.latest_for(order)&.transaction_id || order.number
```
Noah-Silvera, line 57: We should only be aware of transactions that have an `OrderTransaction` associated with them.

If an order has an existing transaction, manually entered by other means external to our extension, there's no way of knowing what it's `transaction_id` would be. i.e. I don't think defaulting to `order.number` is a useful default here.

Noah-Silvera: The first `OrderTransaction` for an order will always be the `order.number` anyways.

Noah-Silvera (changes requested):

app/models/super_good/solidus_taxjar/order_transaction.rb
Noah-Silvera: #115 is up which adds the order transaction! We can rebase this on top of that and use the `transaction_date`

forkata: This change is now rebased over #115 and we have updated the ordering as well as are now setting the `transaction_date` based on the response from Taxjar.

lib/super_good/solidus_taxjar/api.rb
Noah-Silvera: I don't think the fallback is safe, because we haven't considered what we what to do with orders that existed in the app before taxjar.

For example, let's say the following events happen.

1. Someone installs this gem (pre the reporting feature release)
2. Someone then uses this gem to calculate taxes. They then **manually** create transactions in the taxjar web app that correspond to the order.
    * The naming schema they used when manually creating the transactions is completely unknown to us. i.e. `MY-COMPANY-NAME-SOLIDUS-ORDER-NUMBER-28371-RANDOM-CHARACTERS`. We have no guarantee it will be the `order.number`
3. They then update this gem and **enable reporting**, and **don't backfill historical spree orders with their manually entered transactions**
4. They try to refund the above order
    * In that case, this fallback will fail, as we've assumed a default transaction naming schema that is wrong.
    
    
This is a larger problem, that affects all our of reporting API's when they are dealing with pre-existing orders. I've created a issue placeholder to pull out this larger discussion from the scope of this PR. https://github.com/SuperGoodSoft/solidus_taxjar/projects/1#card-71078848

I think for now, we should just throw, as we don't have a good way of handling this case, and we should just move this PR through and figure out how to handle that edge case later.

Noah-Silvera:
Just rebased on master in preparation for a merge

Issue #92, opened 2021-07-23
Conditionally "update" transactions on `order_recalculated` events


benjaminwil:
# What

First, we subscribe to all of the order's  `order_recalculated` events. 

Second, we check if a recalculated order is eligible to re-send a transaction to TaxJar. We need to:

1. Check that the current order is `paid` and `complete`.
2. Determine whether a TaxJar transaction already exists for the current order.
   (Blocked by #89 and #101)

Then, if the transaction meets the above criteria:

- If the `transaction_params(solidus_order)` matches what is in the existing TaxJar transaction:
  - Do nothing.
- If the `transaction_params(solidus_order)` do not match what is in the existing TaxJar transaction:
  - Update the taxjar transaction using the custom refund/create logic #112
- If a transaction doesn't exist:
    - Create a new transaction for the order with the latest amounts.

The top-level of this decision tree can be handled by a service that checks whether TaxJar has the most up-to-date order information. (#106.)

# Why

It's impossible to `PUT` updates to transactions when they're in a tax period that has closed. Posting refunds and new orders makes tax reporting more straightforward for TaxJar end users.

This issue covers the following flows in stock Solidus:

- New or updated line item adjustments
- New or updated order-level adjustments
- New or updated shipment adjustments
- New promotion adjustments
- New reimbursements (refunds or store credit)
- New line item cancellations

We could break down this issue into further sub-tasks, but we thought it was worthwhile to provide the big-picture flow.

Noah-Silvera:
We are going to make the assumption it is safe to assume all paid orders are also complete - we should filter on complete orders.
We should **very clearly** document that **only complete** orders will be pushed to TaxJar

Noah-Silvera:
Idea: we could allow users to configure the states of an order that are checked for it to be created/updated.

Noah-Silvera:
Notes from investigating how "paid" orders work in solidus

The `def paid?` method on the order returns true if the `payment_state` is `paid` **OR** `credit_owed`. We want to ignore `credit_owed` so we can't used `paid?`, we have to directly use the `payment_state`

@alexblackie and I confirmed with investigation that the `payment_state` only goes to paid when the order is **captured** and not just when it is authorized.

forkata:
> The `def paid?` method on the order returns true if the `payment_state` is `paid` **OR** `credit_owed`. We want to ignore `credit_owed` so we can't used `paid?`, we have to directly use the `payment_state`

[...]

PR #111, opened 2021-09-15
Create new transaction when shipped

senemsoy:
What is the goal of this PR?
---
When a shipment is shipped, if there is no transaction for that order, we create a new transaction. The second half of this issue will be completed in https://github.com/SuperGoodSoft/solidus_taxjar/issues/92

---

1. Generate a sandbox app
    * [x] it generates
2. Mark a shipment as shipped
    * [x] creates a new transaction for the order

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

Screenshots
---

### DIFF
--- a/lib/super_good/engine.rb
+++ b/lib/super_good/engine.rb
@@ -5,6 +5,11 @@ module SuperGoodSolidusTaxjar
     isolate_namespace Spree
     engine_name 'super_good_solidus_taxjar'
 
+    if Spree.solidus_gem_version < Gem::Version.new('2.11.0')
+      require root.join('app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber')
+      SuperGood::SolidusTaxjar::Spree::ReportingSubscriber.subscribe!
+    end
+
     include SolidusSupport::EngineExtensions
   end
 end
--- a/lib/super_good/solidus_taxjar.rb
+++ b/lib/super_good/solidus_taxjar.rb
@@ -37,6 +37,10 @@ module SuperGood
       def table_name_prefix
         "solidus_taxjar_"
       end
+
+      def reporting
+        ::SuperGood::SolidusTaxjar::Reporting.new
+      end
     end
 
     self.cache_duration = 3.hours
--- a/lib/super_good/solidus_taxjar.rb
+++ b/lib/super_good/solidus_taxjar.rb
@@ -11,6 +11,7 @@ require "super_good/solidus_taxjar/tax_calculator"
 require "super_good/solidus_taxjar/tax_rate_calculator"
 require "super_good/solidus_taxjar/discount_calculator"
 require "super_good/solidus_taxjar/addresses"
+require "super_good/solidus_taxjar/reporting"
 
 module SuperGood
   module SolidusTaxjar
--- a/lib/super_good/solidus_taxjar/reporting.rb
+++ b/lib/super_good/solidus_taxjar/reporting.rb
@@ -1,0 +1,17 @@
+module SuperGood
+  module SolidusTaxjar
+    class Reporting
+      def initialize(api: SuperGood::SolidusTaxjar.api)
+        @api = api
+      end
+
+      def report_transaction(order)
+        begin
+          @api.show_latest_transaction_for(order)
+        rescue Taxjar::Error::NotFound
+          @api.create_transaction_for(order)
+        end
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - [#120](https://github.com/SuperGoodSoft/solidus_taxjar/pull/120) Change default `SOLIDUS_BRANCH` and `RAILS_VERSION`
 - [#109](https://github.com/SuperGoodSoft/solidus_taxjar/pull/109) Save `OrderTransaction` after API call to TaxJar
 - [#119](https://github.com/SuperGoodSoft/solidus_taxjar/pull/119) Move the install generator into the correct path so that it will be installed in the dummy app.
+- [#111](https://github.com/SuperGoodSoft/solidus_taxjar/pull/111) Create a new taxjar transaction when a shipment is shipped.
 
 ## v0.18.2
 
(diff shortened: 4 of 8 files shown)


benjaminwil (approved):

lib/generators/super_good/solidus_taxjar/install_generator.rb
```diff
@@ -10,6 +10,9 @@ def create_initializer_file
             Spree.config do |config|
               config.tax_calculator_class = SuperGood::SolidusTaxjar::TaxCalculator
             end
+            Spree::Event.subscribe 'shipment_shipped' do |shipment|
+              SuperGood::SolidusTaxjar::Reporting.report_transaction(shipment.order)
+            end
```
forkata, line 15: Is there a reason why we are doing this in an initializer. The standard way is to introduce a subscriber class which can contain all our reporting callbacks https://guides.solidus.io/developers/events/overview.html.

benjaminwil: To expand on Chris's comment: I do think this code needs to be within the extension and not in the app, for extension maintainability reasons.

- We may change the `#report_transaction` method signature someday.
- This may cause errors in the host application if they remove the `solidus_taxjar` extension someday.

senemsoy: @Noah-Silvera can correct me on this, but I think it's to give the users of the extension the option of not opting-in to our new reporting behaviour. If we do this within the extension, it would cause a problem for users who've implemented their own reporting solutions?

benjaminwil: I would argue that if an extension user wanted to implement their own reporting solution they should set `enable_reporting` to `false and then create their own `shipment_shipped` subscriber manually.

I think that our reporting implementation should always work the same way if `enable_reporting` is set to `true`. And I think that means that the logic should live inside of our extension. I worry that asking end users to commit generated code may be confusing and lead to hard-to-debug issues being reported.

lib/super_good/solidus_taxjar.rb
```diff
@@ -36,6 +36,8 @@ def api
 
       def table_name_prefix
         "solidus_taxjar_"
+      def reporting
+        ::SuperGood::SolidusTaxjar::Reporting.new
```
forkata, line 40: Looks like this might be the result of a bad merge but we're missing an `end` tag on the `table_name_prefix` method.

lib/super_good/solidus_taxjar/reporting.rb
```diff
@@ -6,6 +6,7 @@ def initialize(api: SuperGood::SolidusTaxjar.api)
       end
 
       def report_transaction(order)
+        return unless SuperGood::SolidusTaxjar.reporting_enabled
```
forkata, line 9: It would be nice if this was checked inside of the subscriber instead of here. That way people could opt-out of the automated reporting, but still manually call the reporting API. Thoughts on that?

forkata:
@senemsoy I recall we discovered that the install generator was not being run when generating the dummy app due to the improper name-spacing. I think that was something we needed to fix in order for your tests to pass because they depend on the subscriber being called. Based on this https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/lib/spree/testing_support/common_rake.rb we should move the `install_generator.rb` into an `install` folder for that to work in the dummy app.

lib/generators/super_good/solidus_taxjar/install/install_generator.rb
```diff
@@ -12,6 +12,10 @@ def create_initializer_file
             Spree.config do |config|
               config.tax_calculator_class = SuperGood::SolidusTaxjar::TaxCalculator
             end
+            if Spree.solidus_gem_version < Gem::Version.new('2.11.0')
```
Noah-Silvera, line 15: In the spirit of Ben's earlier comment of not putting code that has the power to affect how the reporting feature works in the initializer, we should move this to internal gem code.

One place could be inside the body of the rails engine

````rb
# frozen_string_literal: true

module SuperGoodSolidusTaxjar
  class Engine < Rails::Engine
    isolate_namespace Spree
    engine_name 'super_good_solidus_taxjar'

    if Spree.solidus_gem_version < Gem::Version.new('2.11.0')
      require root.join('app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber')
      SuperGood::SolidusTaxjar::Spree::ReportingSubscriber.subscribe!
    end

    include SolidusSupport::EngineExtensions
  end
end
````

**Review by forkata — APPROVED** [review]

File: lib/super_good/engine.rb [hunk 5]
```diff
@@ -5,6 +5,11 @@ class Engine < Rails::Engine
     isolate_namespace Spree
     engine_name 'super_good_solidus_taxjar'
 
+    if Spree.solidus_gem_version < Gem::Version.new('2.11.0')
+      require root.join('app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber')
+      SuperGood::SolidusTaxjar::Spree::ReportingSubscriber.subscribe!
+    end
```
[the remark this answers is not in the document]
Noah-Silvera: @senemsoy it doesn't look like the `require` line actually got removed!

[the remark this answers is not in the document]
senemsoy: We removed the require for the `Spree::Event::Subscriber` from the reporting subscriber file, as that would be autoloaded. We haven't tested removing the require for the reporting_subscriber in here.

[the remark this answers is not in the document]
senemsoy: We should still require the `reporting_subscriber` here, as it won't be autoloaded unless the app is configured to do that. https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/lib/spree/event/subscriber_registry.rb#L76-L78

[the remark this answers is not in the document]
Noah-Silvera: Ahhh that was my misunderstanding 🤦🏽‍♀️. Thanks for the clarification!

Noah-Silvera (approved):

Noah-Silvera:
Just a rebase on master!

Noah-Silvera:
Just updated the changelog now that we are based on master.

forkata (approved):

benjaminwil (approved):

Noah-Silvera:
@senemsoy thanks for holding out on this one, I know it was open for ages!

senemsoy:
@Noah-Silvera thanks for merging yay!!

Issue #115, opened 2021-10-08
Record transaction date on order transactions


nvandoorn:
We should be recording this now even if we don't use it yet to avoid backfilling.

Noah-Silvera:
#109 completed this issue

PR #116, opened 2021-10-13
[115] Add transaction date to `OrderTransaction`

Noah-Silvera:
What is the goal of this PR?
---

We cannot rely on the `updated_at` and `created_at` value on `OrderTransaction`s to know when the order was originally created in TaxJar. As such, we create a separate column now.

#115 

---

1. Create an order transaction.
    * [x] Ensure that the `transaction_date` is validated
    * [x] Ensure you can create one with a `transaction_date`

Merge Checklist
---

- [x] Run the manual tests
- [x] Rebase on top of #111 to record the transaction date or vice versa
- [x] Update the changelog

### DIFF
--- a/db/migrate/20211008183858_add_transaction_date_to_order_transaction.rb
+++ b/db/migrate/20211008183858_add_transaction_date_to_order_transaction.rb
@@ -1,0 +1,5 @@
+class AddTransactionDateToOrderTransaction < ActiveRecord::Migration[5.0]
+  def change
+    add_column :solidus_taxjar_order_transactions, :transaction_date, :datetime, null: false
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - [#108](https://github.com/SuperGoodSoft/solidus_taxjar/pull/108) Add new model associated with a `Spree::Order` to represent taxjar order creation transactions
 - [#117](https://github.com/SuperGoodSoft/solidus_taxjar/pull/117) Fix migration install
 - [#114](https://github.com/SuperGoodSoft/solidus_taxjar/pull/114) Add new model associated with a `SuperGood::SolidusTaxjar::OrderTransaction` to represent taxjar refund creation transactions
+- [#116](https://github.com/SuperGoodSoft/solidus_taxjar/pull/116) Update the `OrderTransaction` model to record the transaction date.
 
 ## v0.18.2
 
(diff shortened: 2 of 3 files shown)


db/migrate/20211008183858_add_transaction_date_to_order_transaction.rb
```diff
@@ -0,0 +1,5 @@
+class AddTransactionDateToOrderTransaction < ActiveRecord::Migration[5.0]
+  def change
+    add_column :solidus_taxjar_order_transactions, :transaction_date, :datetime, null: false
```
adammathys, line 3: We can't set a `null: false` condition on a new column for an existing table without having some sort of default. (For anyone using this extension that already has transactions, this would make every record invalid.)

We should probably split that out into a separate action and have some strategy for setting it for any existing records. In other words, our migration should do the following:
1. Create new column that can be null.
2. Populate new column on all existing records using an existing value.
3. Alter column to not allow null values.

Steps 2 and 3 could also be contained in a separate migration if desired.

That pattern is also generally a good strategy as it can help avoid locking the table for a long time when there are a lot of records on the table we're adding a column to. It's fast to add a column without a default, but can be very slow if it needs to set a default value for each row as well. (Otherwise the migration could end up taking down a user's store during the deploy.)

adammathys (changes requested):

db/migrate/20211008183858_add_transaction_date_to_order_transaction.rb
Noah-Silvera: Currently, no code has been merged into master that creates order transactions, and there's definitely been no releases with the order transaction model! The code to add the `OrderTransaction` model was merged two weeks ago, and I highly doubt anyone has used it to make their own custom reporting integration yet.

There's not really a sensible default we can provide for this value - we would need to make an API call out to taxjar to get the. correct transaction date! This would lead to quite a complicated migration step to document, as those calls would need to be authenticated.

Although I would normally be all for a safer migration like you are suggesting,  since the chances are extremely low anyone is actually using this model, I think in this case, this migration is acceptable.

What do you think?

adammathys: Extremely low is not zero. 😉

If there's no way for transactions to be created yet in the extension (ignoring anyone who's skipped ahead and implemented it themselves), I'm fine with the migration as is.

For future reference, if the value for a column requires some complicated logic to populate after the fact, you can pick an easily distinguishable one for the migration and provide a rake task that can be run afterwards to clean up the data. (e.g.: Set the initial value to `Time.at(0)` and the go through and update any records with that value.)

adammathys (approved):

swively (approved):

AlistairNorman (approved):

nvandoorn (approved):

PR #127, opened 2021-11-22
Add feature spec for checkout

Noah-Silvera:
What is the goal of this PR?
---

Add a basic acceptance test for the extension so you no longer have to run the painful manual testing step before merging a PR. 

Merge Checklist
---

- [x] Update the changelog
~~- [ ] Run a sandbox app and verify taxes are being calculated~~

### DIFF
--- a/spec/fixtures/cassettes/features/spree/admin/checkout.yml
+++ b/spec/fixtures/cassettes/features/spree/admin/checkout.yml
@@ -1,0 +1,183 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://api.taxjar.com/v2/taxes
+    body:
+      encoding: UTF-8
+      string: '{"to_country":"US","to_zip":"91761","to_city":"Ontario","to_state":"CA","to_street":"450
+        Helen Ave","line_items":[{"id":9999,"quantity":1,"unit_price":"19.99","discount":"-0.0","product_tax_code":"TaxCode
+        - 932150"}],"shipping":"10.0"}'
+    headers:
+      User-Agent:
+      - 'TaxJar/Ruby (Darwin Noahs-MacBook-Pro.local 20.6.0 Darwin Kernel Version
+        20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64
+        x86_64; ruby 2.6.7-p197; OpenSSL 1.1.1k  25 Mar 2021) taxjar-ruby/3.0.2'
+      Authorization:
+      - Bearer <BEARER_TOKEN>
+      X-Api-Version:
+      - '2020-08-07'
+      Plugin:
+      - supergoodsolidustaxjar
+      Connection:
+      - close
+      Content-Type:
+      - application/json; charset=UTF-8
+      Host:
+      - api.taxjar.com
+  response:
+    status:
+      code: 200
+      message: OK
+    headers:
+      Content-Type:
+      - application/json; charset=utf-8
+      Content-Length:
+      - '1116'
+      Connection:
+      - close
+      Date:
+      - Mon, 29 Nov 2021 22:11:40 GMT
+      X-Amzn-Requestid:
+      - da26345e-3a30-4899-bdd9-c72c61f63ee6
+      Access-Control-Allow-Origin:
+      - https://developers.taxjar.com
+      X-Amz-Apigw-Id:
+      - Jll0iHA6oAMF4gw=
+      X-Amzn-Trace-Id:
+      - Root=1-61a5501c-0fad5b964c4a5d3b2b806fe8
+      X-Cache:
+      - Miss from cloudfront
+      Via:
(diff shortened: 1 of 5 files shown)


jarednorman (approved):

benjaminwil:
Something is up with the Capybara configuration for CI:
```
ActionController::RoutingError:
  No route matches [GET] "/fonts.googleapis.com/css"
  ...
```

And I'd be curious to learn why this happened! Very interesting.

spec/fixtures/cassettes/features/spree/admin/checkout.yml
```diff
@@ -0,0 +1,116 @@
...
+      string: '{"to_country":"US","to_zip":"35242","to_city":"Richmond","to_state":"AL","to_street":"143
+        Swan Street","line_items":[{"id":1,"quantity":1,"unit_price":"19.99","discount":"-0.0","product_tax_code":"TaxCode
+        - 593989"}],"shipping":"10.0"}'
+    headers:
+      User-Agent:
+      - 'TaxJar/Ruby (Darwin Noahs-MacBook-Pro.local 20.6.0 Darwin Kernel Version
+        20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64
+        x86_64; ruby 2.6.7-p197; OpenSSL 1.1.1k  25 Mar 2021) taxjar-ruby/3.0.2'
+      Authorization:
+      - Bearer fake-bearer-token
```
benjaminwil, line 17: There is [a way to configure VCR to automatically filter out sensitive information](https://relishapp.com/vcr/vcr/v/5-0-0/docs/configuration/filter-sensitive-data).

I don't see us needing to re-record these request frequently, so maybe this would be overkill for us. But worth noting.

benjaminwil (approved):

spec/fixtures/cassettes/features/spree/admin/checkout.yml
Noah-Silvera: It's a good thing to know about :). May be worth doing! It looks like the CI is failing, so will have to revisit this PR and might tackle that.

forkata: Yes, we can definitely filter out certain headers and other parts of the request/response based on a regex https://relishapp.com/vcr/vcr/docs/configuration/filter-sensitive-data

forkata: This SO seems to provide a good example for how to do this for the `Authorization` header https://stackoverflow.com/a/58629025/1492410

Noah-Silvera:
These tests are failing due to a circle CI problem :( We are just gonna wait a couple days to see if circle CI fixes it.
https://github.com/solidusio/solidus/pull/4211

Noah-Silvera:
Just a rebase on master to try and get the tests passing.

benjaminwil:
Alex and I rebased against `master` and added made a change to the `#fill_in_address` helper method so that it works with Solidus 2.x and 3.x.

Noah-Silvera:
This is the command to reproduce the error on CI (with rails 6.1 and solidus 2.11)

`bundle exec rspec './spec/features/spree/admin/checkout_spec.rb[1:1]' './spec/super_good/solidus_taxjar/api_params_spec.rb[1:6:1]' --seed 1`

Noah-Silvera:
I figured the test error out and I hate it

1. The cassette returns data for a item with `id:1`  and uses that to calculate taxes
2. The `api_params_spec.rb`test that causes the error when run alongside creates a line item first. This causes the `id` counter to move to 2.
4. The line item that the checkout spec creates now has an ID that doesn't match the line item returned in the taxjar cassette.

See the passing tests and last commit of this draft PR https://github.com/SuperGoodSoft/solidus_taxjar/pull/147

Noah-Silvera:
cc: @benjaminwil

benjaminwil:
@Noah-Silvera Can we use VCR request matchers and tell them to ignore the ID part of the request? I think if all other parts of the request are the same, that's still reasonable.

Noah-Silvera:
@benjaminwil it's actually matching ok! 

The problem is that the body of the request that is returned contains a line item with `id:1`
This body is processed by the app, and used to calculate taxes. When the order taxes are calculated in the app, it looks for a line item with `id: 2`, and doesn't find a match in the TaxJar response, so it calculates zero tax for the line item.

One (brittle) fix is to update the response in the cassette to have the `id: 2`

Another would be to intercept the VCR response, and modify it to match whatever the order line item ID is, but I don't know how you'd go about that.

Noah-Silvera:
I fixed the ID in the spec and cassette as discussed in slack.

Noah-Silvera:
Added a comment to the change.

forkata (approved):
:shipit: Thanks Noah, Nick and Co.

Noah-Silvera:
Just changed the `fonts.googleapis` blacklist to `https`!

PR #129, opened 2021-11-29
Report transaction asynchronously

senemsoy:
What is the goal of this PR?
---
To move the reporting of transactions to be performed asynchronously. Previous to this change, when a shipment was shipped, a new reporting transaction was created through API synchronously. This was not ideal because any errors would impact the transition of the shipment to `shipment_shipped`. We have an introduced an Active Job class for reporting, which can be used from other places as well.

---

1. Create an order and mark a shipment as shipped
    * [ ] Assert that a `SuperGood::SolidusTaxjar::ReportTransactionJob` is queued for the order.
2. Perform the job 
    * [ ] Assert that the transaction was reported on Taxjar.

Merge Checklist
---

- [ ] Run the manual tests
- [x] Update the changelog

Screenshots
---

### DIFF
--- a/lib/super_good/solidus_taxjar.rb
+++ b/lib/super_good/solidus_taxjar.rb
@@ -21,6 +21,7 @@ module SuperGood
       attr_accessor :custom_order_params
       attr_accessor :discount_calculator
       attr_accessor :exception_handler
+      attr_accessor :job_queue
       attr_accessor :line_item_tax_label_maker
       attr_accessor :logging_enabled
       attr_accessor :reporting_enabled
@@ -53,6 +54,7 @@ module SuperGood
     self.exception_handler = ->(e) {
       Rails.logger.error "An error occurred while fetching TaxJar tax rates - #{e}: #{e.message}"
     }
+    self.job_queue = :default
     self.line_item_tax_label_maker = ->(taxjar_line_item, spree_line_item) { "Sales Tax" }
     self.logging_enabled = false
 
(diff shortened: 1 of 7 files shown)


adammathys (approved):

Noah-Silvera (approved):
🙌

lib/super_good/solidus_taxjar.rb
```diff
@@ -29,6 +29,7 @@ class << self
       attr_accessor :taxable_address_check
       attr_accessor :taxable_order_check
       attr_accessor :test_mode
+      attr_accessor :job_queue
```
benjaminwil, line 32: The rest of these settings are alphabetized. Maybe we should keep doing that.

lib/super_good/solidus_taxjar.rb
```diff
@@ -66,5 +67,6 @@ def reporting
     self.taxable_address_check = ->(address) { true }
     self.taxable_order_check = ->(order) { true }
     self.test_mode = false
+    self.job_queue = :default
```
benjaminwil, line 70: The rest of these settings are alphabetized. Maybe we should keep doing that.

benjaminwil (approved):

lib/super_good/solidus_taxjar.rb
senemsoy: Done!

forkata (approved):

benjaminwil:
We rebased this against the latest `master` and resolved conflicts.

PR #135, opened 2021-12-07
Run CI build against current versions of Solidus

forkata:
What is the goal of this PR?
---
This removes the lock we have on the Solidus extensions Orb version
which previous to this commit was running test only on "older" versions
of Solidus which were 2.9 and 2.10. With this change to the Orb version
as well as the switch to the "run-tests" command we will be running tests
against master, 3.1, 3.0 and 2.11 which are the officially supported versions
of Solidus currently.

---

1. Create a PR
    * [x] Check that the tests passed on Circle.ci

Merge Checklist
---

- [ ] ~Run the manual tests~
- [x] Update the changelog
- [ ] ~Run a sandbox app and verify taxes are being calculated~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,10 +8,7 @@ jobs:
   run-specs-with-postgres:
     executor: solidusio_extensions/postgres
     steps:
-      - checkout
-      - solidusio_extensions/test-branch:
-          branch: v2.11
-      - solidusio_extensions/store-test-results
+      - solidusio_extensions/run-tests
 
 workflows:
   "Run specs on supported Solidus versions":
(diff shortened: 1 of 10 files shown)


Noah-Silvera:
I'm curious about other's thoughts about maintaining support for solidus 2.9 and solidus 2.10. 

I think it's reasonable to only maintain support against the versions of solidus that are currently supported, and dropping "official" support for 2.9 and 2.10, and tell users who are using those versions to lock this extension to a release that supports them! 

 * We don't have to worry about monkey patching solidus from our gem in case a solidus vulnerability appears that affects 2.9/2.10 and our gem interactions. We can trust solidus to patch it on the supported versions.
 * Reduce the scope, reduce the CI runtime

forkata:
@Noah-Silvera Agree completely with you, I think if we switch to always using the current version of the Circle Orb, we'll ensure that whenever that is updated upstream we also get those changes automatically and always have an eye on how the extension is doing against the supported versions of Solidus.

One other benefit of this change is that we'll ensure new PRs to the repo don't introduce any changes which break against the `master` branch of Solidus, which should hopefully help reduce the maintenance overhead for staying up to date.

Noah-Silvera (approved):

benjaminwil (approved):

PR #138, opened 2021-12-09
Add admin UI for configuring reporting

senemsoy:
What is the goal of this PR?
---
We're now able to enable/disable reporting in the admin UI. The feature is flagged for now, till the reporting functionality is fully implemented. 

---

1. Go to Taxjar settings on the admin panel, turn reporting on, refresh the page.
    * [x] The setting is saved.

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

Screenshots
---

https://user-images.githubusercontent.com/16298232/145460519-30c45363-c9e1-4d0a-a350-42d796fdf876.mov


Noah-Silvera:
You should rebase this on master to run the tests against solidus 2.11! Might fix the failing spec.

spec/features/spree/admin/taxjar_settings_spec.rb
```diff
@@ -22,6 +22,20 @@
       click_on "TaxJar Settings"
     end
 
+    context "Taxjar reporting is enabled" do
+      it "shows that reporting is enabled" do
+        pending("Not implemented yet")
+
+        expect(page).to have_content("Transaction Sync")
+        expect(page).to have_select("Transaction Sync", selected: "No")
+        expect(page).to have_content("Sync orders and refund with TaxJar for automated sales tax reporting and filing. Complete and closed transactions sync automatically on update.")
```
Noah-Silvera, line 30: > "Complete and closed transactions sync automatically on update."

I'd probably roll with "Complete and closed transactions will automatically be synced when they are updated", else I could see the the "on update" part of the message telling someone "Your transactions are automatically backfilled on sync", which is not the case!

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -2,7 +2,20 @@ module Spree
   module Admin
     class TaxjarSettingsController < Spree::Admin::BaseController
       def edit
+        @settings = SuperGood::SolidusTaxjar::Settings.find_or_create_by(id: 1)
       end
+
+      def update
+        settings = SuperGood::SolidusTaxjar::Settings.find_or_create_by(id: 1)
+        settings.update!(settings_params)
+        flash[:alert] = "TaxJar settings updated!"
```
Noah-Silvera, line 11: Is alert a red flash message or green flash message? I didn't see it in the video demo, and I'm used to alerts being red, but this one should probably be green!

Noah-Silvera (commented):
The code and changes look good!

I'm wondering if we should use the [`Spree::Preferences::Configuration`](https://github.com/solidusio/solidus/blob/master/core/lib/spree/preferences/configuration.rb) model? It seems to be an established pattern in solidus that doesn't require creating a new table and managing our own model, and gets rid of the `find_by(id: 1)` queries. It also gives you some styled views and stuff to use and generally seems quite flexible.

Here's the pattern in active shipping!

Controller
https://github.com/solidusio-contrib/solidus_active_shipping/blob/master/app/controllers/spree/admin/active_shipping_settings_controller.rb

View
https://github.com/solidusio-contrib/solidus_active_shipping/blob/master/app/views/spree/admin/active_shipping_settings/edit.html.erb

https://github.com/solidusio-contrib/solidus_active_shipping/blob/master/lib/solidus_active_shipping/configuration.rb

Poking around the solidus-contrib, there's definitely other actively maintained extensions that use it!
https://github.com/solidusio-contrib/solidus_related_products/blob/master/lib/solidus_related_products/configuration.rb
https://github.com/solidusio-contrib/solidus_digital/blob/master/lib/spree/spree_digital_configuration.rb

app/controllers/spree/admin/taxjar_settings_controller.rb
forkata: I think alert is the green one, and error is the red one, but I could be wrong.

forkata: OK, I was wrong, it's actually `:success` we should be using based on the style guide here https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/style_guide/topics/messaging/_flashes.html.erb
0

app/models/super_good/solidus_taxjar/settings.rb
```diff
@@ -0,0 +1,6 @@
+module SuperGood
+  module SolidusTaxjar
+    class Settings < ActiveRecord::Base
```
benjaminwil, line 3: We had a conversation IRL about maybe using Solidus's built-in preferences stores to keep TaxJar extension settings in the DB. Has any more thought been put into that since our conversation?

(I'm guessing no. Everyone's been pretty busy.)

benjaminwil (commented):
This seems good.

benjaminwil:
> I'm wondering if we should use the `Spree::Preferences::Configuration` model?

This seems like a good idea.

This settings UI is going to affect all of the extension's future settings UI and maybe requires a group discussion. 

Is this PR trying to do too much? We could creating a settings UI that does nothing before we start porting over static configuration settings to be end-user configurable ones.

nvandoorn:
Since last round of reviews:

- The model layer has been reworked to use `Spree::Base` and `Spree::Preferences::Persistable` to be consistent with other Solidus preferences
- We use partials provided by Solidus in the UI
- Re-based against master

spec/features/spree/admin/taxjar_settings_spec.rb
nvandoorn: Had a quick look at this and decided we may want to update the copy after speaking with Ryan. For now we keep the copy exactly the same as the Magento integration.

nvandoorn:
Latest revision re-bases onto master and uses the `success` flash instead of `alert`

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -2,6 +2,20 @@ module Spree
   module Admin
     class TaxjarSettingsController < Spree::Admin::BaseController
       def edit
+        @configuration = SuperGood::SolidusTaxjar.configuration
+      end
+
+      def update
+        configuration = SuperGood::SolidusTaxjar.configuration
+        configuration.update!(configuration_params)
+        flash[:success] = "TaxJar settings updated!"
```
forkata, line 11: Should we use `update` here and test for the result before showing the flash message? If the update fails now it'll return a 500 error to the user.

spec/requests/spree/admin/taxjar_settings_request_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'spec_helper'
+
+RSpec.describe 'Admin TaxJar Settings', :type => :request do
+  # https://github.com/solidusio-contrib/solidus_subscriptions/blob/master/spec/controllers/spree/admin/subscriptions_controller_spec.rb
```
forkata, line 4: I am not sure we care about referencing where we got this code from.

spec/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber_spec.rb
```diff
@@ -26,11 +26,7 @@
     end
 
     context "reporting is disabled" do
-      before do
-        allow(::SuperGood::SolidusTaxjar)
-          .to receive(:reporting_enabled)
-          .and_return(false)
-      end
+      let(:reporting_enabled) { false }
```
forkata, line 29: I don't think this line is doing anything right now because we only reference this in the `before` for the parallel context above. I believe we're just getting the default value of `false` here when a new configuration is created the first time it's accessed.

forkata (commented):
Thanks @nvandoorn! I've left a few small comments otherwise this looks good to me!

forkata:
We should probably add a Changelog entry to document the move of the `reporting_enabled` flag to the admin as well as the new `reporting_ui_enabled` configuration option.

benjaminwil (approved):

nvandoorn:
Since last review:

- Changelog entry added
- Use `update` instead of `update!` to avoid showing admin users a 500
- Small spec cleanup suggested by Chris

adammathys (approved):

forkata (approved):

benjaminwil (approved):

PR #161, opened 2022-02-09
Register `shipment_shipped` event

nvandoorn:
What is the goal of this PR?
---

Our CI is failing on Solidus master, so we try to fix it by registering the `shipment_shipped` event. I don't think this event is in the right place yet, so looking for feedback on where others think it belongs.

### DIFF
--- a/app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
+++ b/app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
@@ -3,6 +3,10 @@ module SuperGood
     module Spree
       module ReportingSubscriber
         include ::Spree::Event::Subscriber
+
+        if ::Spree::Event.method_defined?(:register)
+          ::Spree::Event.register("shipment_shipped")
+        end
         event_action :report_transaction, event_name: :shipment_shipped
 
         def report_transaction(event)


adammathys (approved):

Noah-Silvera:
Green specs on solidus master ????? what is this magic?   😍

jarednorman (approved):

senemsoy (approved):

forkata:
> I don't think this event is in the right place yet, so looking for feedback on where others think it belongs.
Yeah, we may want to open up an issue to Solidus and raise this. Based on the original discussion on the PR that introduced the registry it was mentioned that this should be done in an initializer, however when we were testing that approach we realized that the subscribers are loaded by the solidus support engine module before initializers are run.

Noah-Silvera:
@nvandoorn  @forkata what if we did this in the engine?

There's already a precedent for this, as we register subscribers for < 2.11 in the engine!

https://github.com/SuperGoodSoft/solidus_taxjar/blob/49991319f5fef71623881cac810036f25b3ec481/lib/super_good/engine.rb#L8

forkata:
@Noah-Silvera I think my main concern was around possible issues due to code reloading causing the event to be registered more than once which I think would raise an error. I don't know if putting it in the engine is better than in the subscriber or if that will behave differently with code reloading. IRRC we confirmed that works as well.

I feel like the recommended way to register events based on [this comment](https://github.com/solidusio/solidus/pull/4226#issuecomment-1001972214) possibly did not considered how that may be done by extensions, so I feel like having a discussion around that may be productive.

Noah-Silvera:
> @Noah-Silvera I think my main concern was around possible issues due to code reloading causing the event to be registered more than once which I think would raise an error. I don't know if putting it in the engine is better than in the subscriber or if that will behave differently with code reloading. IRRC we confirmed that works as well.
> 
> I feel like the recommended way to register events based on [this comment](https://github.com/solidusio/solidus/pull/4226#issuecomment-1001972214) possibly did not considered how that may be done by extensions, so I feel like having a discussion around that may be productive.

Ahhh yes! I see. Super down for this approach for now and having a larger discussion in solidus around how to handle this with extensions! Can re-evaluate based on the outcome of that.

Noah-Silvera (approved):

PR #169, opened 2022-02-23
Add button to backfill transactions

nvandoorn:
What is the goal of this PR?
---

Basic implementation of transaction backfill with naive logging

---

1. Visit the TaxJar settings admin page
2. Click "Backfill transactions"
3. Verify all  _complete_ transactions have a corresponding record on your TaxJar account

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

Screenshots
---

Screen Shot 2022-03-21 at 14 37 59PDT
Screen Shot 2022-03-21 at 14 37 05PDT

### DIFF
--- a/app/controllers/spree/admin/taxjar_settings_controller.rb
+++ b/app/controllers/spree/admin/taxjar_settings_controller.rb
@@ -31,6 +31,12 @@ module Spree
         redirect_back(fallback_location: spree.admin_taxjar_settings_path)
       end
 
+      def backfill_transactions
+        ::SuperGood::SolidusTaxjar::BackfillTransactions.call
+        flash[:success] = "Transactions Backfilled!"
+        redirect_back(fallback_location: edit_admin_taxjar_settings_path)
+      end
+
       private
 
       def configuration_params
--- a/app/controllers/spree/admin/taxjar_settings_controller.rb
+++ b/app/controllers/spree/admin/taxjar_settings_controller.rb
@@ -32,7 +32,7 @@ module Spree
       end
 
       def backfill_transactions
-        ::SuperGood::SolidusTaxjar::BackfillTransactions.call
+        ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
         flash[:success] = "Transactions Backfilled!"
         redirect_back(fallback_location: edit_admin_taxjar_settings_path)
       end
--- a/app/controllers/spree/admin/taxjar_settings_controller.rb
+++ b/app/controllers/spree/admin/taxjar_settings_controller.rb
@@ -32,9 +32,8 @@ module Spree
       end
 
       def backfill_transactions
-        ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
-        flash[:success] = "Transactions Backfilled!"
-        redirect_back(fallback_location: edit_admin_taxjar_settings_path)
+        @backfilled_order_numbers = ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
+        flash[:success] = "Successfully backfilled transactions for #{@backfilled_order_numbers.count} orders."
       end
 
       private
--- a/lib/super_good/solidus_taxjar/backfill_transactions.rb
+++ b/lib/super_good/solidus_taxjar/backfill_transactions.rb
@@ -1,8 +1,23 @@
 module SuperGood
   module SolidusTaxjar
     class BackfillTransactions
-      def self.call
+      def initialize(api: SuperGood::SolidusTaxjar.api)
+        @api = api
       end
+
+      def call
+        backfilled_orders = []
+        ::Spree::Order.complete.where(shipment_state: 'shipped').find_each do |order|
+          next if order.taxjar_order_transactions.any?
+          api.create_transaction_for(order)
+          backfilled_orders.push(order.number)
+        end
+        backfilled_orders
+      end
+
+      private
+
+      attr_reader :api
     end
   end
 end
--- a/lib/super_good/solidus_taxjar/backfill_transactions.rb
+++ b/lib/super_good/solidus_taxjar/backfill_transactions.rb
@@ -1,0 +1,8 @@
+module SuperGood
+  module SolidusTaxjar
+    class BackfillTransactions
+      def self.call
+      end
+    end
+  end
+end
--- a/spec/features/spree/admin/taxjar_settings_spec.rb
+++ b/spec/features/spree/admin/taxjar_settings_spec.rb
@@ -42,7 +42,6 @@ RSpec.feature 'Admin TaxJar Settings', js: true, vcr: true do
       scenario "the user backfills their transactions" do
         visit "/admin/taxjar_settings/edit"
         click_on "Backfill Transactions"
-        pending "basic feedback on what happened with the backfill is implemented"
         expect(page).to have_content("Successfully backfilled transactions for 1 orders.")
         expect(page).to have_content(order.number)
       end
--- a/spec/features/spree/admin/taxjar_settings_spec.rb
+++ b/spec/features/spree/admin/taxjar_settings_spec.rb
@@ -23,6 +23,23 @@ RSpec.feature 'Admin TaxJar Settings', js: true, vcr: true do
       click_on "TaxJar Settings"
     end
 
+    context "order is shipped" do
+      let(:order) { create :shipped_order }
+
+      before do
+        create(:tax_rate, name: "Sales Tax")
+        # The order must be created **after** the TaxRate to have the correct totals
+        order
+      end
+
+      scenario "the user backfills their transactions" do
+        click_on "Backfill Transactions"
+        pending "basic feedback on what happened with the backfill is implemented"
+        expect(page).to have_content("Successfully backfilled transactions for 1 orders.")
+        expect(page).to have_content(order.number)
+      end
+    end
+
     context "Taxjar reporting is enabled" do
       it "shows that reporting is enabled" do
         expect(page).to have_content("Transaction Sync")
--- a/spec/features/spree/admin/taxjar_settings_spec.rb
+++ b/spec/features/spree/admin/taxjar_settings_spec.rb
@@ -14,13 +14,20 @@ RSpec.feature 'Admin TaxJar Settings', js: true, vcr: true do
       allow(ENV).to receive(:[]).and_call_original
       allow(ENV).to receive(:[]).with("TAXJAR_API_KEY").and_return(api_token)
       allow(SuperGood::SolidusTaxjar).to receive(:reporting_ui_enabled).and_return(true)
+    end
 
+    scenario "the user navigates to the TaxJar Settings" do
       visit "/admin"
       click_on "Settings"
       expect(page).to have_content("Taxes")
       click_on "Taxes"
       expect(page).to have_content("TaxJar Settings")
       click_on "TaxJar Settings"
+
+      within('[data-hook="admin_settings_taxes_tabs"] > .active') do
+        expect(page).to have_content("TaxJar Settings")
+      end
+      expect(page).to have_content("Transaction Sync")
     end
 
     context "order is shipped" do
@@ -33,6 +40,7 @@ RSpec.feature 'Admin TaxJar Settings', js: true, vcr: true do
       end
 
       scenario "the user backfills their transactions" do
+        visit "/admin/taxjar_settings/edit"
         click_on "Backfill Transactions"
         pending "basic feedback on what happened with the backfill is implemented"
         expect(page).to have_content("Successfully backfilled transactions for 1 orders.")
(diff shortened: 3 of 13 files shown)


Noah-Silvera (approved):

spec/features/spree/admin/taxjar_settings_spec.rb
```diff
@@ -23,6 +24,13 @@
       click_on "TaxJar Settings"
     end
 
+    it "the user backfills their transactions" do
```
AlistairNorman: With capybara `scenario` is an alias of `it` that makes these kinds of tests read better.

lib/super_good/solidus_taxjar/backfill_transactions.rb
```diff
@@ -0,0 +1,8 @@
+module SuperGood
```
AlistairNorman, line 1: I would generally make this class and actually implement it in the commit before this one. This commit is already quite long.

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -1,10 +1,6 @@
 module Spree
   module Admin
     class TaxjarSettingsController < Spree::Admin::BaseController
-      def edit
```
AlistairNorman, line 4: Why is this being removed? It's not mentioned in the commit message.

spec/features/spree/admin/taxjar_settings_spec.rb
```diff
@@ -9,6 +9,7 @@
 
   describe "Taxjar settings tab" do
     let(:api_token) { "token" }
+    let(:order) { create :order }
```
AlistairNorman: I don't think this variable is referenced.

spec/features/spree/admin/taxjar_settings_spec.rb
```diff
@@ -24,10 +22,9 @@
       click_on "TaxJar Settings"
     end
 
-    it "the user backfills their transactions" do
+    it "the user backfills their transactions", vcr: {cassette_name: "features/spree/admin/backfill"} do
       click_on "Backfill Transactions"
       expect(page).to have_content("Transactions Backfilled!")
-      pending "basic feedback on what happened with the backfill is implemented"
       expect(page).to have_content("#{order.number} successfully backfilled.")
```
AlistairNorman, line 28: This isn't what you made the logs say is it?

app/controllers/spree/admin/taxjar_backfills_controller.rb
```diff
@@ -2,9 +2,9 @@ module Spree
   module Admin
     class TaxjarBackfillsController < Spree::Admin::BaseController
       def create
-        ::SuperGood::SolidusTaxjar::BackfillTransactions.call
-        flash[:success] = "Transactions Backfilled!"
```
AlistairNorman: I think there was a test for this wasn't there? We probably don't want to be adding this line with a test if we're going to delete it in the same PR.

lib/super_good/solidus_taxjar/backfill_transactions.rb
```diff
@@ -2,6 +2,13 @@ module SuperGood
   module SolidusTaxjar
     class BackfillTransactions
       def self.call
+        backfilled_orders = []
+        ::Spree::Order.complete.find_each do |order|
+          next if order.taxjar_order_transactions.any?
```
AlistairNorman, line 7: I haven't looked at what this association is but could we get the database to do this instead of looping over them all and skipping the ones that have transactions?

spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      example.run
+      ::SuperGood::SolidusTaxjar.test_mode = false
+    end
+
+    before do
+      reported_order = create :completed_order_with_totals
+      create(:taxjar_order_transaction, order: reported_order)
+      create :order
+
+      allow(::SuperGood::SolidusTaxjar).to receive(:api).and_return(api_spy)
```
AlistairNorman, line 21: We could get rid of this if we injected the dependency instead of having it be implicit.

spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+RSpec.describe SuperGood::SolidusTaxjar::BackfillTransactions do
+  describe ".call" do
+    subject { described_class.call }
+
+    let!(:complete_order) { create :completed_order_with_totals }
+    let(:api_spy) { instance_spy(::SuperGood::SolidusTaxjar::Api) }
+
+    around do |example|
+      ::SuperGood::SolidusTaxjar.test_mode = true
```
AlistairNorman, line 11: I'm not sure but is this necessary? Looks like we're stubbing out the api call anyways.

lib/super_good/solidus_taxjar/backfill_transactions.rb
```diff
@@ -2,6 +2,13 @@ module SuperGood
...
+          SuperGood::SolidusTaxjar.api.create_transaction_for(order)
+          backfilled_orders.push(order.number)
+        end
+        backfilled_orders
```
AlistairNorman, line 11: If we found a way to query for the orders we could pass in the orders that need to be backfilled and we wouldn't need to have this return anything. We could loosen our coupling a bit.

AlistairNorman (changes requested):
Some of these comments are more just style but I think a few small changes here could go a long way.

spec/features/spree/admin/taxjar_settings_spec.rb
Noah-Silvera: It's referenced here!

https://github.com/SuperGoodSoft/solidus_taxjar/blob/8689f6c871f7c52baf86e30f403f7caa78529c33/spec/features/spree/admin/taxjar_settings_spec.rb#L31

app/controllers/spree/admin/taxjar_backfills_controller.rb
Noah-Silvera: The implementation changed over the course of the PR as the feature developed. Initially, a flash message was sufficient communication, but then as more logs were added, displaying them on the page itself was needed.

This line isn't the final implementation, but it's part of the history of building up the implementation.

spec/features/spree/admin/taxjar_settings_spec.rb
Noah-Silvera: Fixed!

app/controllers/spree/admin/taxjar_settings_controller.rb
Noah-Silvera: Added a note to the commit message!

spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
Noah-Silvera: Updated.

lib/super_good/solidus_taxjar/backfill_transactions.rb
benjaminwil: I could imagine a more efficient query where we only grab orders where the `COUNT` of TaxJar order transactions is `0`, but this is a "run it once" kind of process, so probably not critical to optimize it.

benjaminwil: It might be worth noting that orders are completed before they are shipped. So not all complete orders are liable for tax.

Noah-Silvera: @benjaminwil good catch! Updated to also check for an orders ship state.

lib/super_good/solidus_taxjar/backfill_transactions.rb
AlistairNorman: It would loosen the coupling because currently this receives an order from the controller and then returns the orders. Passing the orders would be looser because it wouldn't return anything so the coupling is only one direction. This isn't a blocker or anything but removing this return would be more of a Tell don't Ask implementation.

app/controllers/spree/admin/taxjar_backfills_controller.rb
AlistairNorman: Hmm, I'm not sure if this is accurate but it seems like things were committed when they were written as opposed to when it would make the most sense for someone reading. i.e. committing an empty service class earlier and committing one behaviour then changing it later. I can see how this might happen but I don't think it's ideal for either the reviewers or anyone looking back at this history later.

AlistairNorman (approved):

spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      example.run
+      ::SuperGood::SolidusTaxjar.test_mode = false
+    end
+
+    before do
+      reported_order = create :shipped_order
+      create(:taxjar_order_transaction, order: reported_order)
+      create :completed_order_with_totals
+      create :order_ready_to_ship
+      create :order
```
adammathys, line 21: I don't think we need to create all of these orders. We can't have a shipped incomplete order so just a single order here should suffice to test that condition.

spec/features/spree/admin/taxjar_settings_spec.rb
```diff
@@ -15,16 +15,21 @@
...
+    end
 
+    scenario "the user navigates the the TaxJar Settings" do
       visit "/admin"
       click_on "Settings"
       expect(page).to have_content("Taxes")
       click_on "Taxes"
       expect(page).to have_content("TaxJar Settings")
       click_on "TaxJar Settings"
+      expect(page).to have_content("TaxJar Settings")
```
adammathys, line 25: Might be better to have a different assertion here. The click could fail and this will still pass because that text is on the previous page.

app/controllers/spree/admin/taxjar_backfills_controller.rb
```diff
@@ -2,9 +2,9 @@ module Spree
   module Admin
     class TaxjarBackfillsController < Spree::Admin::BaseController
       def create
-        ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
-        flash[:success] = "Transactions Backfilled!"
-        redirect_back(fallback_location: edit_admin_taxjar_settings_path)
+        order_numbers_backfilled = ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
+        @logs = "Transactions Backfilled! #{order_numbers_backfilled.map { |number| "#{number} successfully backfilled." }.join(" ")}"
```
adammathys, line 6: I know this is just a first draft, but I think we can make this messaging a little nicer. Maybe something like:
```ruby
"Successfully backfilled transactions for #{order_numbers_backfilled.count} orders. (#{order_numbers_backfilled.join(", ")})"
```

No need to print out `successfully backfilled` for every single order.

benjaminwil (approved):

app/controllers/spree/admin/taxjar_backfills_controller.rb
Noah-Silvera: Addressed!

spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
Noah-Silvera: Addressed!

spec/requests/spree/admin/taxjar_backfill_request_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+
+RSpec.describe 'Admin TaxJar Backfill', type: :request do
+  # https://github.com/solidusio-contrib/solidus_subscriptions/blob/master/spec/controllers/spree/admin/subscriptions_controller_spec.rb
```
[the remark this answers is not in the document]
Noah-Silvera: Deleted!

spec/features/spree/admin/taxjar_settings_spec.rb
Noah-Silvera: Added an assertion to check that the tab is active and for some basic content of the tab!

Noah-Silvera:
This last push just addressed a comment around a better assertion in `taxjar_settings_spec.rb` feature spec. That is the only file that changed.

Noah-Silvera:
This last push fixed most of the broken tests - there's just one or two more that need fixing.

Noah-Silvera:
Just looking for a re-review as there's been some signficant changes in regards to how the logs are displayed

AlistairNorman (approved):

forkata (approved):

benjaminwil (approved):

Noah-Silvera:
Tested and updated the changelog! Setting auto-merge

PR #175, opened 2022-03-04
Add request logging to TaxJar API requests

nvandoorn:
What is the goal of this PR?
---

Add request logging!

---

1. Perform an action that will call the TaxJar API (e.g checkout so taxes are calculated)
2. Inspect the logs for requests to TaxJar (e.g grep for the host)

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

### DIFF
--- a/lib/super_good/solidus_taxjar.rb
+++ b/lib/super_good/solidus_taxjar.rb
@@ -62,7 +62,7 @@ module SuperGood
     self.custom_order_params = ->(order) { {} }
     self.discount_calculator = ::SuperGood::SolidusTaxjar::DiscountCalculator
     self.exception_handler = ->(e) {
-      Rails.logger.error "An error occurred while fetching TaxJar tax rates - #{e}: #{e.message}"
+      self.logger.error "An error occurred while fetching TaxJar tax rates - #{e}: #{e.message}"
     }
     self.job_queue = :default
     self.line_item_tax_label_maker = ->(taxjar_line_item, spree_line_item) { "Sales Tax" }
--- a/lib/super_good/solidus_taxjar.rb
+++ b/lib/super_good/solidus_taxjar.rb
@@ -1,6 +1,7 @@
 require "solidus_core"
 require "solidus_support"
 require "taxjar"
+require "super_good/solidus_taxjar/overrides/request_override"
 
 require "super_good/solidus_taxjar/version"
 require "super_good/solidus_taxjar/transaction_id_generator"
--- a/lib/super_good/solidus_taxjar.rb
+++ b/lib/super_good/solidus_taxjar.rb
@@ -66,7 +66,7 @@ module SuperGood
     }
     self.job_queue = :default
     self.line_item_tax_label_maker = ->(taxjar_line_item, spree_line_item) { "Sales Tax" }
-    self.logging_enabled = false
+    self.logging_enabled = true
 
     # The reporting setting in the admin UI is disabled for now till the reporting
     # feature is fully implemented.
--- a/lib/super_good/solidus_taxjar.rb
+++ b/lib/super_good/solidus_taxjar.rb
@@ -31,6 +31,7 @@ module SuperGood
       attr_accessor :taxable_address_check
       attr_accessor :taxable_order_check
       attr_accessor :test_mode
+      attr_writer :logger
 
       def configuration
         ::SuperGood::SolidusTaxjar::Configuration.default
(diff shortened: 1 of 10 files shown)


lib/super_good/solidus_taxjar/overrides/request_override.rb
```diff
@@ -0,0 +1,15 @@
+module SuperGood
```
Noah-Silvera, line 1: Maybe we namespace this under `Taxjar::Api` also?
I don't have a super strong opinion on that.

lib/super_good/solidus_taxjar/api.rb
```diff
@@ -25,7 +25,7 @@ def tax_for(order)
         taxjar_client.tax_for_order(ApiParams.order_params(order)).tap do |taxes|
           next unless SuperGood::SolidusTaxjar.logging_enabled
 
-          Rails.logger.info(
```
Noah-Silvera, line 28: These logs are very verbose - I'm not sold on enabling them by default! 

The fact that the `logging_enabled` variable was set to `false` by default makes me think that `logging_enabled` was intended to enable **debug** logs.

Could we hide these behind a debug logging flag or just log them with `debug` instead of `info` ?

Noah-Silvera (changes requested):

lib/super_good/solidus_taxjar/api.rb
forkata: You are raising a good point here, should we just remove this in favour of the `debug` output which will contain the full response body? This seems redundant in the context of the logging we're adding in this PR 🤔

Noah-Silvera: If the logging added in this PR doesn't just log the response status and endpoint, then yeah, I think this would be redundant and removable!

If that's the case, could we add some tests for logging the response body?

nvandoorn: There is a test to match the body but it's a little bit hard to spot: https://github.com/SuperGoodSoft/solidus_taxjar/pull/175/files#diff-47df37a1fa490b55235d448f0a525a2f00e8d3e696588594f52fe20f7b0991feR26

We can't match on too much of the body because the response may change over time.

nvandoorn: We also decided to remove the inline log statement. It's completely redundant with request logging in place.

nvandoorn:
Latest revision:
- Remove an extra log statement that isn't needed
- Add documentation on configuring logger
- Updated the exception handler to use the configured logger
- CHANGELOG

lib/super_good/solidus_taxjar/api.rb
Noah-Silvera: Ahh cool! Sounds great!

Noah-Silvera (approved):

lib/super_good/solidus_taxjar.rb
```diff
@@ -31,6 +31,7 @@ class << self
       attr_accessor :taxable_address_check
       attr_accessor :taxable_order_check
       attr_accessor :test_mode
+      attr_accessor :logger
```
AlistairNorman, line 34: This should be `attr_writer` I think. I guess this technically works but you're defining a `logger` method and then immediately overriding it.

AlistairNorman (approved):
Just a couple minor comments. I feel a little uneasy about that commit with no spec changes. It's probably safe but I still don't love seeing changes to untested code.

lib/super_good/solidus_taxjar/api.rb
```diff
@@ -22,13 +22,7 @@ def tax_categories
...
 
       def tax_for(order)
-        taxjar_client.tax_for_order(ApiParams.order_params(order)).tap do |taxes|
-          next unless SuperGood::SolidusTaxjar.logging_enabled
-
-          Rails.logger.info(
-            "TaxJar response for #{order.number}: #{taxes.to_h.inspect}"
-          )
-        end
+        taxjar_client.tax_for_order(ApiParams.order_params(order))
```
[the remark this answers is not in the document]
forkata: There were no tests for the logging that was removed here!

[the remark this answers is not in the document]
nvandoorn: Just confirmed that there is no specs for these log calls. I suspect they were added in a rush.

nvandoorn:
Latest revision:

- Fix the specs (global state leaking between examples)
- Make the `logger` attribute and `attr_writer` as per Alistair's feedback.

PR #179, opened 2022-03-18
Provide valid state and zip in address factory

nvandoorn:
What is the goal of this PR?
---
The zip code and state code provided by the default
factory were not a valid pair. This is fine for most tests, but causes
issues when making requests to TaxJar. To fix this, we provide a valid
New York zipcode.

---

- [ ] Write a test that requires recording a TaxJar request with an address
- [ ] Verify the request doesn't error because of the zip/state

### DIFF
--- a/lib/super_good/solidus_taxjar/testing_support/factories/address_factory.rb
+++ b/lib/super_good/solidus_taxjar/testing_support/factories/address_factory.rb
@@ -1,0 +1,11 @@
+FactoryBot.modify do
+  # Solidus's default address factories provide an invalid state and zipcode
+  # combination.
+  factory :address do
+    zipcode { "11430" }
+
+    transient do
+      state_code { "NY" }
+    end
+  end
+end


Noah-Silvera (approved):
🙏

lib/super_good/solidus_taxjar/testing_support/factories/address_factory.rb
```diff
@@ -0,0 +1,12 @@
+FactoryBot.modify do
+  # Solidus's default address factories provide an invalid state and zipcode
+  # combination.
+  factory :address do
+    zipcode { "11430" }
+
+    transient do
+
```
Noah-Silvera, line 8: Little nit, we could delete this whitespace.

forkata (approved):
Thanks @benjaminwil and @nvandoorn!

AlistairNorman (approved):
It's not 90210 but I suppose this zip works 😄

benjaminwil (approved):
Thanks for doing this!

adammathys (approved):

PR #181, opened 2022-03-23
Take adjustments into account in discount calculator

Noah-Silvera:
What is the goal of this PR?
---

TaxJar requires the line item totals to match the order total when posting transactions. 

We need to ensure any adjustments that are not tax are included in the line item discount.

Merge Checklist
---

- [x] Run the manual tests
- [x] Run the scenario in https://github.com/SuperGoodSoft/solidus_taxjar/issues/184 and ensure it is fixed.
- [x] Update the changelog

### DIFF
--- a/lib/super_good/solidus_taxjar/discount_calculator.rb
+++ b/lib/super_good/solidus_taxjar/discount_calculator.rb
@@ -6,7 +6,7 @@ module SuperGood
       end
 
       def discount
-        -line_item.promo_total
+        -1 * line_item.adjustments.select { |value| !value.tax? && value.eligible? }.sum(&:amount)
       end
 
       private
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - [#174](https://github.com/SuperGoodSoft/solidus_taxjar/pull/174) Provide a link to Taxjar state settings
 - [#177](https://github.com/SuperGoodSoft/solidus_taxjar/pull/177) Make nexus caching configurable
 - [#169](https://github.com/SuperGoodSoft/solidus_taxjar/pull/169) Add basic backfill transaction functionality
+- [#181](https://github.com/SuperGoodSoft/solidus_taxjar/pull/181) Take all non-tax adjustment types into account when calculating a line item's discount
 
 ## v0.18.2
 
--- a/spec/super_good/solidus_taxjar/api_params_spec.rb
+++ b/spec/super_good/solidus_taxjar/api_params_spec.rb
@@ -139,6 +139,20 @@ RSpec.describe SuperGood::SolidusTaxjar::ApiParams do
   describe ".order_params" do
     subject { described_class.order_params(order) }
 
+    before do
+      # The discount calculator relies on the line item adjustments existing in
+      # order to calculate the correct discount amount for TaxJar.
+      create(
+        :adjustment,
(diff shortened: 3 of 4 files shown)


lib/super_good/solidus_taxjar/discount_calculator.rb
```diff
@@ -6,7 +6,7 @@ def initialize(line_item)
       end
 
       def discount
-        -line_item.promo_total
+        -(line_item.promo_total + line_item.adjustments.select { |value| !value.tax? && value.eligible? }.sum(&:amount))
```
forkata, line 9: Me and @nvandoorn expanded on the tests added here and discovered that this will double count promotion adjustments since they are already included in the `promo_total`. We have an updated version of this change which completely removes the `promo_total` and just does the sum of the adjustments in Ruby which may be closer to what we want.

forkata (changes requested):

jarednorman (approved):

Noah-Silvera:
Updated to take into account all non-tax adjustment types in the discount calculation!

senemsoy (approved):

Noah-Silvera:
Just added a test for non-eligible adjustments as we were missing one!

lib/super_good/solidus_taxjar/discount_calculator.rb
```diff
@@ -6,7 +6,7 @@ def initialize(line_item)
       end
 
       def discount
-        -line_item.promo_total
+        line_item.adjustments.select { |value| !value.tax? && value.eligible? }.sum(&:amount)
```
nvandoorn, line 9:
```ruby
line_item.adjustments.non_tax.eligible.sum(:amount)
```

nvandoorn (commented):
Small suggestion that came up during today's mob session. This will save us from filtering in Ruby code and use SQL queries to find the appropriate adjustments.

lib/super_good/solidus_taxjar/discount_calculator.rb
Noah-Silvera: Not going to implement this, as discussed in the weekly checkin, as we want this calculator to be able to handle in memory orders.

Noah-Silvera:
Just a rebase on master!

@forkata just need another review from you as changes were requested!

nvandoorn (approved):

forkata (approved):
It looks like we need to update some specs here to take into account the new calculation we are doing, but otherwise this looks good to me, thanks @Noah-Silvera!

Noah-Silvera:
Drafting this for now

In order to get the `api_params_spec` working, a major refactor to the spec to actually use the calculated values from adjustments and promotions on the order needed to be undertaken.

This work is not done yet - I've pushed a WIP which recalculates promotions correctly but tax is going to be a bigger problem. 

There was also a bug where the sign of the value in the discount calculator was flipped

benjaminwil (approved):
Beautiful !! :chefs-kiss:

forkata (approved):
Pending this fixes our issues when re-calculating tax after a `UnitCancel` adjustment is created, this looks good to me 👍🏼 Thanks @Noah-Silvera and Co.

nvandoorn:
Gave this a test and it does resolve the issue described in #184

nvandoorn (approved):

PR #182, opened 2022-03-26
Supply a default tax rate

Noah-Silvera:
What is the goal of this PR?
---

All tax adjustments need a TaxRate as a source. Previously, we were relying on a user to have a tax rate with the name "Sales Tax" configured. This was an easy to miss configuration step. We should instead create the placeholder `Spree::TaxRate` desired with a descriptive name.

---

1. Calculate some taxes
2. Visit the tax rates page of the admin
    * [x] Ensure the default rate is shown

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@
 - [#177](https://github.com/SuperGoodSoft/solidus_taxjar/pull/177) Make nexus caching configurable
 - [#169](https://github.com/SuperGoodSoft/solidus_taxjar/pull/169) Add basic backfill transaction functionality
 - [#181](https://github.com/SuperGoodSoft/solidus_taxjar/pull/181) Take all non-tax adjustment types into account when calculating a line item's discount
+- [#182](https://github.com/SuperGoodSoft/solidus_taxjar/pull/182) Automatically create default Tax Rate
+
+## Upgrading Instructions
+
+* If you had previously configured a `Spree::TaxRate` with the name "Sales Tax", it can be deleted after upgrading, as a new `Spree::TaxRate` with the name "Solidus TaxJar Rate" will automatically be created. Alternatively, you can rename your existing `Spree::TaxRate` from "Sales Tax" to "Solidus TaxJar Rate". [#182](https://github.com/SuperGoodSoft/solidus_taxjar/pull/182)
 
 ## v0.18.2
 
--- a/README.md
+++ b/README.md
@@ -48,9 +48,7 @@ introduced in Solidus v2.4. This maps better to how the TaxJar API itself works.
     [Configuration](#configuration).
 
 1. Finally, make sure that the `TAXJAR_API_KEY` environment variable is set to
-  your TaxJar API key and make sure that you have a `Spree::TaxRate` with the name
-  "Sales Tax". This will be used as the source for the tax adjustments that
-  Solidus creates.
+  your TaxJar API key.
 
 ## Project Status
 
--- a/README.md
+++ b/README.md
@@ -84,6 +84,8 @@ Solidus < 2.11, so this extension provides a [compatibility layer](app/overrides
 breakdown for a given `Spree::Order`. The breakdown includes separate line items
 taxes and shipment taxes.
 
+This tax calculator will create a `Spree::TaxRate` that is required for tax adjustments. All other `Spree::TaxRate`s will be ignored in calculations. See [this wiki page](https://github.com/SuperGoodSoft/solidus_taxjar/wiki/How-Solidus-TaxJar-calculates-tax) for more details.
+
 ### TaxRateCalculator
 
 `SuperGood::SolidusTaxjar::TaxRateCalculator` allows calculating the tax rate
--- a/lib/super_good/solidus_taxjar/tax_calculator.rb
+++ b/lib/super_good/solidus_taxjar/tax_calculator.rb
@@ -109,8 +109,14 @@ module SuperGood
         )
       end
 
+      # Tax adjustments require an associated Tax Rate. This tax rate is not
+      # used in the extension's calculations as the rates from the TaxJar API
+      # are used, so a placeholder rate and calculator is created.
       def tax_rate
-        ::Spree::TaxRate.find_by(name: "Sales Tax")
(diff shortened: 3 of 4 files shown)


lib/super_good/solidus_taxjar/tax_calculator.rb
```diff
@@ -109,8 +109,14 @@ def no_tax
...
       end
 
+      # Tax adjustments require an associated Tax Rate. This tax rate is not
+      # used in the extension's calculations as the rates from the TaxJar API
+      # are used, so a placeholder rate and calculator is created.
       def tax_rate
-        ::Spree::TaxRate.find_by(name: "Sales Tax")
+        ::Spree::TaxRate.find_or_create_by!(name: "TaxJar Placeholder Rate") do |tax_rate|
+          tax_rate.calculator = ::Spree::Calculator.new
+          tax_rate.amount = 0
```
AlistairNorman, line 118: What's the purpose of this line? I noticed it wasn't tested so I was wondering if it's important it should maybe have a test to go along with the change, if it's not important do we need it at all?

AlistairNorman (approved):
Are you going to get a bunch of failures because you made the new tax amount 0 and the old one was 0.5?

lib/super_good/solidus_taxjar/tax_calculator.rb
Noah-Silvera: The amount is not used at all as the rate is just used as a placeholder and the real tax rate is calculated by the TaxJar API

I was basically trying to setup the properties of the rate in the most generic way possible to signify it is a placeholder rate and doesn't have a real impact, and setting the amount to `0` seemed like a good value to communicate that!

senemsoy (approved):

adammathys (approved):

benjaminwil (commented):
I think this is great, but I'm curious how you decided on the name "TaxJar Placeholder Rate".

I think this name, too, like the previous name, could be confusing. Someone might decide, "Oh, a placeholder? I should delete this."

Maybe a more explicit name, like "Solidus TaxJar Connection Rate", could be advantageous.

forkata (commented):
Thanks @Noah-Silvera and @adammathys! I am in favour of this change in general, I am just not sure if we need to change the name of the tax rate.

forkata:
> The motivation for changing the name is that currently users will end up with a rate in their system with a default amount of 0.05 and no indication that it is connected to our extension other than a little piece of documentation in the readme.

I am not sure that is the case outside of specs. We currently rely on users setting up the tax rate themselves, so the amount on the rate would be whatever the user creates, right?

I think I am onboard with the change proposed here. I do think we should use a slightly better name, something that non-technical users can recognize and possibly also leave the note in the documentation that the rate must exists, or it will be created by the extension. I like @benjaminwil's proposed name but would drop the "Connection" part of it as that doesn't really add any clarity. Something like "Solidus TaxJar Rate" or just "TaxJar Rate" makes the most sense to me.

README.md
```diff
@@ -48,9 +48,7 @@ introduced in Solidus v2.4. This maps better to how the TaxJar API itself works.
     [Configuration](#configuration).
 
 1. Finally, make sure that the `TAXJAR_API_KEY` environment variable is set to
-  your TaxJar API key and make sure that you have a `Spree::TaxRate` with the name
-  "Sales Tax". This will be used as the source for the tax adjustments that
-  Solidus creates.
```
forkata, line 53: Could we change this to something like 
```
The extension looks for a `Spree::TaxRate` with the name "<NAME>", and if one is not found it will be created. This will be used as the source for the tax adjustments that Solidus creates.
```
I think documenting what tax rate the extension uses in the README is valuable information.

Noah-Silvera: I'm gonna make a wiki page regarding tax rates and move that information there in an effort to consolidate new documentation there!

Noah-Silvera:
> I am not sure that is the case outside of specs. We currently rely on users setting up the tax rate themselves, so the amount on the rate would be whatever the user creates, right?

We discussed this on slack, but good to make a note of it here
This extension completely replaces the calculator that calculates tax based on `Spree::TaxRate`s. No matter how many `Spree::TaxRate`s with whatever percentage exist, the taxjar API rates will still be used. Going to make a wiki page describing how this happens in more detail and I'll link it here!

Noah-Silvera:
I added a wiki page better describing how this extension hooks into tax calculation in solidus!

https://github.com/SuperGoodSoft/solidus_taxjar/wiki/How-Solidus-TaxJar-calculates-tax

Noah-Silvera:
Added a link to the wiki page in the readme and updated the changelog

Noah-Silvera:
Just a rebase on master!

benjaminwil (approved):

CHANGELOG.md
```diff
@@ -33,6 +33,11 @@
 - [#174](https://github.com/SuperGoodSoft/solidus_taxjar/pull/174) Provide a link to Taxjar state settings
 - [#177](https://github.com/SuperGoodSoft/solidus_taxjar/pull/177) Make nexus caching configurable
 - [#169](https://github.com/SuperGoodSoft/solidus_taxjar/pull/169) Add basic backfill transaction functionality
+- [#182](https://github.com/SuperGoodSoft/solidus_taxjar/pull/182) Automatically create default Tax Rate
+
+## Upgrading Instructions
+
+* If you had previously configured a `Spree::TaxRate` with the name "Sales Tax", it can be deleted after upgrading. [#182](https://github.com/SuperGoodSoft/solidus_taxjar/pull/182)
```
forkata, line 40: Could we also suggest the option of renaming the "Sales Tax" calculator to "Solidus TaxJar Rate" if users don't want to end up with a _new_ calculator after this upgrade. 
I know on the project I work we have specs where we explicitly setup this tax rate when we want to get actual taxes through TaxJar which will also have to be updated.

Noah-Silvera: I like that idea!

Noah-Silvera: Updated!

forkata (approved):

forkata:
Looks like tests are failing due to this https://github.com/octokit/octokit.rb/pull/1420

Noah-Silvera:
Just pushed to fix a changelog conflict.

PR #183, opened 2022-04-13
Allow admin to use TaxJar tax codes

senemsoy:
What is the goal of this PR?
---
Meet certification requirements for assigning products to a TaxJar tax category.

Merge Checklist
---

- [x] Run the manual tests (no need as it's covered in the feature spec)
- [x] Update the changelog

Screenshots
---
Screen Shot 2022-04-13 at 11 12 14 AM
Screen Shot 2022-04-13 at 11 12 25 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 - [#198](https://github.com/SuperGoodSoft/solidus_taxjar/pull/198) Don't create zero dollar transactions when an order is fully refunded.
 - [#195](https://github.com/SuperGoodSoft/solidus_taxjar/pull/195) Run transaction backfills asynchronously
 - [#200](https://github.com/SuperGoodSoft/solidus_taxjar/pull/200) Expand solidus gem into dependencies to support Solidus 3.2+
+- [#183](https://github.com/SuperGoodSoft/solidus_taxjar/pull/183) Add ability to fetch tax codes from TaxJar and assign to tax categories in admin
 
 ## Upgrading Instructions
 
--- a/app/controllers/spree/admin/taxjar_settings_controller.rb
+++ b/app/controllers/spree/admin/taxjar_settings_controller.rb
@@ -16,25 +16,40 @@ module Spree
       end
 
       def sync_nexus_regions
-        if ENV["TAXJAR_API_KEY"]
-          begin
-            Rails.cache.write(
-              :nexus_regions,
-              SuperGood::SolidusTaxjar.api.nexus_regions,
-              expires_in: SuperGood::SolidusTaxjar.cache_duration
-            )
-            flash[:success] = "Updated with new Nexus Regions"
-          rescue Taxjar::Error => exception
-            flash[:error] = exception.message
-          end
+        api_sync do
+          Rails.cache.write(
+            :nexus_regions,
+            SuperGood::SolidusTaxjar.api.nexus_regions,
+            expires_in: SuperGood::SolidusTaxjar.cache_duration
+          )
+          flash[:success] = "Updated with new Nexus Regions"
         end
-        redirect_back(fallback_location: spree.admin_taxjar_settings_path)
+      end
+
+      def sync_tax_categories
+        api_sync do
+          Rails.cache.write(
+            :tax_categories,
+            SuperGood::SolidusTaxjar.api.tax_categories,
+            expires_in: 30.days
+          )
+          flash[:success] = "Updated with new tax categories"
+        end
       end
 
       private
 
       def configuration_params
         params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled)
+      end
+
+      def api_sync
+        begin
+          yield
+        rescue Taxjar::Error => exception
+          flash[:error] = "Failed to complete request to TaxJar: #{exception.message}"
+        end
+        redirect_back(fallback_location: spree.admin_taxjar_settings_path)
       end
     end
   end
--- a/app/controllers/spree/admin/taxjar_settings_controller.rb
+++ b/app/controllers/spree/admin/taxjar_settings_controller.rb
@@ -1,17 +1,17 @@
 module Spree
   module Admin
     class TaxjarSettingsController < Spree::Admin::BaseController
+      helper_method :taxjar_tax_category_option
+
       def edit
         @configuration = SuperGood::SolidusTaxjar.configuration
         @nexus_regions = Rails.cache.fetch(:nexus_regions) || []
+        @tax_categories = cached_tax_categories
       end
 
       def update
-        if SuperGood::SolidusTaxjar.configuration.update(configuration_params)
-          flash[:success] = "TaxJar settings updated!"
-        else
-          flash[:alert] = "Failed to update settings!"
-        end
+        configuration_update if params[:super_good_solidus_taxjar_configuration]
+        tax_category_update if params[:tax_category]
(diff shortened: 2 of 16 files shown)


app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -41,6 +47,17 @@ def backfill_transactions
       def configuration_params
         params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled)
       end
+
+      def api_sync
+        if ENV["TAXJAR_API_KEY"]
+          begin
+            yield
+          rescue Taxjar::Error => exception
+            flash[:error] = exception.message
```
forkata, line 56: Should we provide some prefix to this message, like "Failed to complete request to TaxJar: ...`. The exception message here might not contain anything meaningful to an end user.

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -41,6 +47,17 @@ def backfill_transactions
       def configuration_params
         params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled)
       end
+
+      def api_sync
+        if ENV["TAXJAR_API_KEY"]
```
forkata, line 52: Is there any reason why we're guarding for this here 🤔 I know we have a similar check around the buttons in the view, but are we worried that someone is going to post directly to the endpoint?

I think sprinkling these checks throughout all our code is a bit of an anti-pattern. I would rather we let the request fail than keep adding checks like this.

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -58,6 +58,36 @@ def api_sync
         end
         redirect_back(fallback_location: spree.admin_taxjar_settings_path)
       end
+
+      def cache_tax_categories
```
forkata, line 62: Would `cached_tax_categories` be a better name here, since we rely on this returning the tax categories?

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -1,17 +1,17 @@
...
       end
 
       def update
-        if SuperGood::SolidusTaxjar.configuration.update(configuration_params)
-          flash[:success] = "TaxJar settings updated!"
-        else
-          flash[:alert] = "Failed to update settings!"
-        end
+        configuration_update if params[:super_good_solidus_taxjar_configuration]
+        tax_category_update if params[:tax_category]
```
forkata, line 14: I am not sure how I feel about reusing the same action for both updates here. Is there a reason why we don't want to add a new controller or even a separate action?

This is probably a bigger conversation but we've already started using this controller for various things unrelated to taxjar settings and I would love if we can stop adding more things to it and start splitting the responsibilities into separate controllers.

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -1,17 +1,17 @@
 module Spree
   module Admin
     class TaxjarSettingsController < Spree::Admin::BaseController
+      helper_method :taxjar_tax_category_option
+
       def edit
         @configuration = SuperGood::SolidusTaxjar.configuration
         @nexus_regions = Rails.cache.fetch(:nexus_regions) || []
+        @tax_categories = cache_tax_categories
```
forkata, line 9: If we have a button to pull down the tax categories on demand, do we still want to try and do it on load as well?

forkata (commented):
Thanks @senemsoy and @benjaminwil! I left a few comments here mostly relating to the decision to keep adding things to the same edit/update action on the settings controller. I feel like this is getting a bit hard to maintain as we add more functionality and I am wondering if this isn't a good time to start splitting these responsibilities into a separate controller or at least separate actions.

app/controllers/spree/admin/taxjar_settings_controller.rb
benjaminwil: Yeah, good suggestion.

app/controllers/spree/admin/taxjar_settings_controller.rb
benjaminwil: I think from a UX perspective, it is great to have all of these Solidus TaxJar settings on a single page.

I think that having more, non-RESTful controller actions would be better overall. But I think that may be an unpopular opinion.

forkata: We can still have the settings on a single page and delegate the actions to a different controller though? If we are worried about keeping things RESTful then we've already muddied the waters by not having a separate controller for the resources we're trying to manage here.

senemsoy: We're interested in making this change Chris, but maybe in another PR? So that we can go ahead with merging the working functionality.

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -56,6 +56,35 @@ def api_sync
         end
         redirect_back(fallback_location: spree.admin_taxjar_settings_path)
       end
+
+      def cached_tax_categories
+        Rails.cache.fetch(
+          :tax_categories,
+          expires_in: 30.days
```
forkata, line 63: Should we use the `SuperGood::SolidusTaxjar.cache_duration` here as well?

app/views/spree/admin/taxjar_settings/edit.html.erb
```diff
@@ -5,27 +5,70 @@
...
+        <td><%= region.country %></td>
+        <td><%= region.region_code %></td>
+        <td><%= region.region %></td>
+      </tr>
+    <% end %>
+    </table>
+    <a class="btn btn-primary" href="<%= spree.admin_taxjar_settings_sync_nexus_regions_path %>">Sync Nexus Regions</a>
+    <a class="btn btn-outline-primary" href="https://app.taxjar.com/account#states">Go to TaxJar to configure states</a>
+  </fieldset>
+  <fieldset>
```
forkata, line 31: We didn't do this for the previous section (nexus regions), but do you mind pulling this into a partial so it's easier to move around in the future 🙏🏼

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 - [#169](https://github.com/SuperGoodSoft/solidus_taxjar/pull/169) Add basic backfill transaction functionality
 - [#181](https://github.com/SuperGoodSoft/solidus_taxjar/pull/181) Take all non-tax adjustment types into account when calculating a line item's discount
 - [#182](https://github.com/SuperGoodSoft/solidus_taxjar/pull/182) Automatically create default Tax Rate
+- [#183](https://github.com/SuperGoodSoft/solidus_taxjar/pull/183) Allow admin to use TaxJar tax codes
```
forkata, line 40: Could we re-phrase this to be a bit more end-user friendly. Maybe something like `Add ability to fetch tax codes from TaxJar and assign to tax categories in admin`

forkata (approved):
Thanks for addressing all the comments here, this looks good to me. I left a couple nit comments but feel free to merge as-is.

app/controllers/spree/admin/taxjar_settings_controller.rb
senemsoy: The default cache duration is 2 hours, but in this case we don't want to update the tax categories this frequently. Users can manually refresh the tax categories.

benjaminwil (approved):

PR #185, opened 2022-04-21
Backfill transactions in batches

nvandoorn:
---

1. Verify you have some orders that are ready for back-filling (e.g paid and shipped)
2. Verify the back-fill feature is enabled
3. Request a manual back-fill
4. Open a Rails console on the sandbox app to verify the new models were created (`TransactionSyncBatch` and `TransactionSyncLog`)

Merge Checklist
---

- [x] Update the changelog
- [ ] Run the manual tests

### DIFF
--- a/app/controllers/spree/admin/taxjar_settings_controller.rb
+++ b/app/controllers/spree/admin/taxjar_settings_controller.rb
@@ -32,8 +32,8 @@ module Spree
       end
 
       def backfill_transactions
-        @backfilled_order_numbers = ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
-        flash[:success] = "Successfully backfilled transactions for #{@backfilled_order_numbers.count} orders."
+        @transaction_sync_batch = ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
+        flash[:success] = "Queued transaction backfill for #{@transaction_sync_batch.transaction_sync_logs.count} orders."
       end
 
       private
--- a/spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
+++ b/spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
@@ -5,7 +5,9 @@ RSpec.describe SuperGood::SolidusTaxjar::BackfillTransactions do
     subject { described_class.new(api: api_spy).call }
 
     let!(:shipped_order) { create :shipped_order }
+    let(:reporting_mock) { instance_double ::SuperGood::SolidusTaxjar::Reporting }
     let(:api_spy) { instance_spy(::SuperGood::SolidusTaxjar::Api) }
+    let(:test_transaction_id) { "R1234-transaction" }
 
     around do |example|
       ::SuperGood::SolidusTaxjar.test_mode = true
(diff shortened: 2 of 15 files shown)


benjaminwil (approved):

spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
```diff
@@ -17,16 +19,58 @@
       reported_order = create :shipped_order
       create(:taxjar_order_transaction, order: reported_order)
       create :order_ready_to_ship
-    end
 
-    it "reports all completed and shipped orders to taxjar" do
```
Noah-Silvera, line 22: I don't think this test should have been deleted!

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -32,8 +32,8 @@ def sync_nexus_regions
       end
 
       def backfill_transactions
-        @backfilled_order_numbers = ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
-        flash[:success] = "Successfully backfilled transactions for #{@backfilled_order_numbers.count} orders."
+        @transaction_sync_batch = ::SuperGood::SolidusTaxjar::BackfillTransactions.new.call
+        flash[:success] = "Successfully backfilled transactions for #{@transaction_sync_batch.transaction_sync_logs.count} orders."
```
Noah-Silvera, line 36: Just wanted to note that since now we are actually handling failures to report transactions, it's possible this message will be incorrect as not all of the logs may correspond to successful transactions. Totally fine if we want to handle that in a future PR!

Noah-Silvera (commented):
Looks great! Just requesting changes for the spec that I don't think should be deleted!

Noah-Silvera (dismissed):
forgot to request changes

Noah-Silvera:
Updated to fix my own requested changes

app/controllers/spree/admin/taxjar_settings_controller.rb
forkata: Thats a good callout @Noah-Silvera! We can easily split this up based on the status to errored/succeeded counts. As you mentioned we're already working on the batching UI, so we'll have another PR shortly to redirect to the new views where the user can see the detailed logs.

forkata (approved):

app/controllers/spree/admin/taxjar_settings_controller.rb
forkata: We changed the copy here in preparation for moving this to an async job in a future PR. Let me know if this makes more sense in terms of the language.

benjaminwil (approved):

Noah-Silvera:
Tried to test this but ended up stuck because a line item on the order we were trying to backfill had two sales tax adjustments - one which had a `source_type` of `Spree::TaxRate` and another with a `source_type` of `nil`.

We suspect this may have just been a problem with the local environment or the demo app on the solidus/rails combo versions that were being run (solidus 3 and rails 6)

nvandoorn:
Tested this again on my machine with expected results!

PR #188, opened 2022-05-10
Show transaction sync batches in user interface

nvandoorn:
What is the goal of this PR?
---

Show the user the status of transaction sync batches

---

1. Sync some orders to TaxJar
1. Visit `/admin/transaction_sync_batches` and verify a batch with a status is present

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

Screenshots
---

Screen Shot 2022-05-10 at 2 37 04 PM

### DIFF
--- a/app/controllers/spree/admin/transaction_sync_batches_controller.rb
+++ b/app/controllers/spree/admin/transaction_sync_batches_controller.rb
@@ -1,0 +1,9 @@
+module Spree
+  module Admin
+    class TransactionSyncBatchesController < Spree::Admin::BaseController
+      def index
+        @batches = SuperGood::SolidusTaxjar::TransactionSyncBatch.all
+      end
+    end
+  end
+end
--- a/app/controllers/spree/admin/transaction_sync_batches_controller.rb
+++ b/app/controllers/spree/admin/transaction_sync_batches_controller.rb
@@ -2,7 +2,7 @@ module Spree
   module Admin
     class TransactionSyncBatchesController < Spree::Admin::BaseController
       def index
-        @batches = SuperGood::SolidusTaxjar::TransactionSyncBatch.all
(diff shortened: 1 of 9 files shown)


app/models/super_good/solidus_taxjar/transaction_sync_batch.rb
```diff
@@ -1,3 +1,9 @@
 class SuperGood::SolidusTaxjar::TransactionSyncBatch < ApplicationRecord
   has_many :transaction_sync_logs
+
+  def status
+    return 'processing' if transaction_sync_logs.processing.any?
+    return 'error' if transaction_sync_logs.error.any?
+    return 'success' if transaction_sync_logs.count == transaction_sync_logs.success.count
```
adammathys, line 7: Should we include a default state should all of the conditions fail? It looks like we're relying on this method returning a string, because if it doesn't the call below (`batch.status.capitalize`) will throw an error.

adammathys (dismissed):

nvandoorn:
Latest push adds a default case to the `#status` method on transaction sync batches.

app/controllers/spree/admin/transaction_sync_batches_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Spree
+  module Admin
+    class TransactionSyncBatchesController < Spree::Admin::BaseController
+      def index
+        @batches = SuperGood::SolidusTaxjar::TransactionSyncBatch.all
```
forkata, line 5: We should probably add some pagination here 🤔

nvandoorn:
Latest revision adds a new commit to add pagination!

Noah-Silvera:
@nvandoorn think this just needs a rebase now that the base branch is merged!

Noah-Silvera (commented):
Last 3 commits look good pending a changelog update!

Noah-Silvera (approved):

PR #189, opened 2022-05-11
Configure `postgres` user on CircleCI

nvandoorn:
What is the goal of this PR?
---

Fix the CI!

There is a recent change to `solidusio/circleci-orbs-extensions` that
updates the postgres/ruby docker image which seems to have broken most
postgres based workflows:

https://github.com/solidusio/circleci-orbs-extensions/commit/d5370b9855dc5bc97542d40e059dccab210d83cc

The specs are failing because the role "postgres" does not exist, so
there is a good chance we don't have the correct user configured. The
commit mentioned above does not change the user, just the postgres/ruby
image version, so it's strange to see this break now. Maybe the old
ruby/postgres images ignored these settings?

This commit pulls the "postgres executor" configuration into this file
so we can test changing the user from "root" to "postgres". If this
change works, we should upstream this to
`solidusio/circleci-orbs-extensions`.

---

- [ ] CI Green

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Upstream the change


benjaminwil:
I'm not opposed to merging this with your forked version of `solidus_dev_support` until the changes are upstreamed.

.circleci/config.yml
```diff
@@ -6,14 +6,32 @@ orbs:
...
+    docker:
+      - image: cimg/ruby:2.7-browsers
+        environment:
+          DB: postgresql
+          PGHOST: 127.0.0.1
+          PGUSER: postgres
+          RAILS_ENV: test
+      - image: cimg/postgres:14.2
+        environment:
+          POSTGRES_USER: postgres
```
forkata, line 36: Could we dry this up by defining this yaml in our own executor block as per this documentation https://circleci.com/docs/2.0/reusing-config/#authoring-reusable-executors

Gemfile
```diff
@@ -22,6 +22,7 @@ group :development, :test do
   gem "pry"
   gem "pry-stack_explorer"
   gem "pry-byebug"
+  gem "solidus_dev_support", github: "nvandoorn/solidus_dev_support"
```
forkata, line 25: I took a brief look at the changes you have here and I am curious if this is something we can override in our extensions' spec helper while we PR this upstream 🤔

forkata (commented):
Thanks for figuring this out @nvandoorn! I left a couple of comments but otherwise, nice work here.

nvandoorn:
Latest revision DRY's up the code a bit by defining an executor and then inlines some Capybara config rather than forking solidus_dev_support.

forkata (approved):
Thanks @nvandoorn

PR #193, opened 2022-06-24
Update extension for new Omnes event 🚌 

forkata:
What is the goal of this PR?
---
Get tests passing against Solidus `master`.
Update the way we use the Solidus event bus to be forwards compatible with the new Omens bus implementation that is now default in Solidus "=> 3.2.0.alpha" - https://github.com/solidusio/solidus/pull/4342

Additional information on updating to the new event bus implementation - https://github.com/solidusio/solidus/discussions/4380

How do you manually test these changes?
---

1. Test the reporting features against Solidus `master`
    * [ ] Create and ship an order and verify it is created as a transaction on TaxJar
    * [ ] Update an order (cancel an item) and confirm the original transaction is refunded on TaxJar and a replacement is created
1. Run the same tests against Solidus 3.0

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

### DIFF
--- a/app/overrides/super_good/solidus_taxjar/spree/order_updater/fire_recalculated_event.rb
+++ b/app/overrides/super_good/solidus_taxjar/spree/order_updater/fire_recalculated_event.rb
@@ -4,6 +4,10 @@ module SuperGood
   module SolidusTaxjar
     module Spree
       module OrderUpdater
+        # Responsible for introducing the `order_recalculated` event for older
+        # versions of Solidus (from 2.9 up until but not including 2.11) which
+        # do not have this event. TaxJar transaction reporting relies on this
+        # event firing when an order is recalculated.
         module FireRecalculatedEvent
           def persist_totals
             ::Spree::Event.fire 'order_recalculated', order: order
--- a/app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
+++ b/app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
@@ -3,6 +3,7 @@ module SuperGood
     module Spree
       module ReportingSubscriber
         include ::Spree::Event::Subscriber
+        include SolidusSupport::LegacyEventCompat::Subscriber
 
         if ::Spree::Event.method_defined?(:register)
           ::Spree::Event.register("shipment_shipped")
--- a/super_good-solidus_taxjar.gemspec
+++ b/super_good-solidus_taxjar.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "solidus_core", ">= 2.4.0"
-  spec.add_dependency "solidus_support"
+  spec.add_dependency "solidus_support", ">= 0.9.0"
   spec.add_dependency "taxjar-ruby"
 
   spec.add_development_dependency "solidus_dev_support"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master
 
+- [#193](https://github.com/SuperGoodSoft/solidus_taxjar/pull/193) Bump version requirement for `solidus_support` to ">= 0.9.0", and as a result drop support for Rails 5.2.
 - [#190](https://github.com/SuperGoodSoft/solidus_taxjar/pull/190) Add transaction sync batch show page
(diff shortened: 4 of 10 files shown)


app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
```diff
@@ -3,9 +3,12 @@ module SolidusTaxjar
     module Spree
       module ReportingSubscriber
         include ::Spree::Event::Subscriber
+        include SolidusSupport::LegacyEventCompat::Subscriber
 
         if ::Spree::Event.method_defined?(:register)
           ::Spree::Event.register("shipment_shipped")
+        elsif ("::Spree::Bus".constantize rescue nil)
```
forkata, line 10: I don't like this, and we may be able to do this in a better way. One thing that would be cool is if the `SolidusSupport::LegacyEventCompat::Bus` class had a `#register` method. Right now it only supports `#publish`.

benjaminwil:
```suggestion
        elsif class_available?("::Spree::Bus")
        
# ...

def class_available?(class_name)
  # ...
end
```

Maybe not worth doing, but moving the constantize check into a private method would make the control flow easier to skim.

benjaminwil (approved):

app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
forkata: After reading through the [update instructions](https://github.com/solidusio/solidus/pull/4342) on the PR that added Omnes, I think the better way to register this is in the initializer which is already conditional on the Solidus verison.

app/overrides/super_good/solidus_taxjar/spree/order_updater/fire_recalculated_event.rb
```diff
@@ -4,6 +4,10 @@ module SuperGood
   module SolidusTaxjar
     module Spree
       module OrderUpdater
+        # Responsible for introducing the `order_recalculated` event for older
+        # versions of Solidus (from 2.9 up until but not including 2.11) which
+        # do not have this event. TaxJar transaction reporting relies on this
+        # event firing when an order is recalculated.
         module FireRecalculatedEvent
```
forkata, line 11: After some extensive testing with @stewart we discovered it may be time to drop this backport as the current state of the extension is no longer compatible with Solidus < 2.11 due to us referencing the persistable preferences module which was introduced in that version [here](https://github.com/SuperGoodSoft/solidus_taxjar/blob/master/app/models/super_good/solidus_taxjar/configuration.rb#L1).

Noah-Silvera: I'm on board with dropping support as we don't support Solidus 2.10 anymore!

forkata: I'll throw a task for this on the board!

super_good-solidus_taxjar.gemspec
```diff
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "solidus_core", ">= 2.4.0"
-  spec.add_dependency "solidus_support"
+  spec.add_dependency "solidus_support", ">= 0.9.0"
```
Noah-Silvera, line 29: This will also force us to drop support for Rails 5.2, which seems ok as it's also officially unsupported now

Noah-Silvera (approved):

super_good-solidus_taxjar.gemspec
forkata: Hmm, I should probably add a changelog entry for this, thanks for calling it out!

nvandoorn (approved):

forkata:
Converting this to a draft as me and @Noah-Silvera discovered an issue with the `shipment_shipped` event not firing the event action. Additionally it looks like with the latest changes on Solidus `master` to remove `solidus_frontend` there are some issues with spinning up the sandbox and dummy apps, so the green checkmark here is stale and will disappear on a rebuild 😞

forkata:
@Noah-Silvera We confirmed the issue with Solidus 3.2 is present upstream as well so not related to the changes merged here. I will work on manually setting up a working sandbox on that version of Solidus so we can verify the new behaviour introduced here.

PR #195, opened 2022-06-27
[150] Asynchronously process transaction backfills

Noah-Silvera:
What is the goal of this PR?
---

Transaction backfills have the potential to encompass a large number of orders, so we should execute them asychronously, and direct the user to the in progress batch display page.

---

* [x] Smoke test backfilling transactions

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

### DIFF
--- a/app/jobs/super_good/solidus_taxjar/backfill_transaction_sync_batch_job.rb
+++ b/app/jobs/super_good/solidus_taxjar/backfill_transaction_sync_batch_job.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module SuperGood
+  module SolidusTaxjar
+    class BackfillTransactionSyncBatchJob < ApplicationJob
+      queue_as { SuperGood::SolidusTaxjar.job_queue }
+
+      def perform(transaction_sync_batch)
+        ::Spree::Order.complete.where(shipment_state: 'shipped').find_each do |order|
+          next if order.taxjar_order_transactions.any?
+          transaction_sync_log = SuperGood::SolidusTaxjar::TransactionSyncLog.create!(
+            transaction_sync_batch: transaction_sync_batch,
+            order: order
+          )
+          begin
+            order_transaction = SuperGood::SolidusTaxjar.reporting.show_or_create_transaction(order)
+            transaction_sync_log.update!(order_transaction: order_transaction, status: :success)
+          rescue Taxjar::Error => exception
+            transaction_sync_log.update!(status: :error, error_message: exception.message)
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
+++ b/spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
@@ -1,7 +1,7 @@
 require "spec_helper"
 
 RSpec.describe SuperGood::SolidusTaxjar::BackfillTransactions do
-  describe ".call" do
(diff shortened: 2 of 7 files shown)


app/jobs/super_good/solidus_taxjar/backfill_transaction_sync_batch_job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module SuperGood
+  module SolidusTaxjar
+    class BackfillTransactionSyncBatchJob < ApplicationJob
+      queue_as { SuperGood::SolidusTaxjar.job_queue }
+
+      def perform(transaction_sync_batch)
+        ::Spree::Order.complete.where(shipment_state: 'shipped').find_each do |order|
```
forkata, line 9: I am trying to think if there is a more forward compatible way to do the order query 🤔 We know we are going to be adding filtering at the batch level, so what do you think about pre-building the `TransactionSyncLog` records on the batch before queueing the job and fetching the orders through those records 🤔

forkata (commented):
Looks like we may need to update a couple of specs here as a result of moving this process to async.

spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
```diff
@@ -2,11 +2,10 @@
 
 RSpec.describe SuperGood::SolidusTaxjar::BackfillTransactions do
   describe ".call" do
```
AlistairNorman, line 4: I know it's not in this PR but this should be `describe "#call" do` because it's an instance method. That being said since we aren't passing anything to the initializer anymore it could also just be a class method instead.

AlistairNorman (approved):

Noah-Silvera:
Sorry just a rebase on master since the base branch was merged! No changes yet

app/jobs/super_good/solidus_taxjar/backfill_transaction_sync_batch_job.rb
Noah-Silvera: I'm hesitant to do much in the ways of forward compatible! We are adding date filtering at the batch level, but what will that look like?

Maybe it looks like a "start_time" and "end_time" column on the `transaction_sync_batch`, in which case this query could just be modified to be like `::Spree::Order.complete.where(shipment_state: 'shipped', created_at: [transaction_sync_batch.start_time..transaction_sync_batch.end_time])` which seems pretty forward compatible to me! But there could be another implementation, and I'd be hesitant to try and make this work better for an implementation we don't know the shape of yet!

Maybe I'm misunderstanding what you are meaning by filtering

spec/super_good/solidus_taxjar/backfill_transactions_spec.rb
Noah-Silvera: Addressed!

app/jobs/super_good/solidus_taxjar/backfill_transaction_sync_batch_job.rb
forkata: It's been so long I already forgot what I was going for here 😂 Let's leave this for now as is and we can refactor when the right time comes 👍🏼

forkata (approved):

forkata:
Tests still need some attention, but otherwise :shipit:

Noah-Silvera:
Just a push to fix the tests!

PR #199, opened 2022-09-07
[150] Render transaction sync log error message

Noah-Silvera:
What is the goal of this PR?
---

When an order fails to sync in a transaction backfill, we should render the reason in the UI.

---

1. Backfill a transaction and have it fail
    * [x] Ensure that the error message is displayed in the UI

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

Screenshots
---
Screen Shot 2022-09-07 at 14 31 07PDT

### DIFF
--- a/app/views/spree/admin/transaction_sync_batches/show.html.erb
+++ b/app/views/spree/admin/transaction_sync_batches/show.html.erb
@@ -11,6 +11,7 @@
       <th>Order</th>
       <th>TaxJar Transaction ID</th>
       <th>Status</th>
+      <th>Error Message</th>
       <th>Created at</th>
       <th>Updated at</th>
     </tr>
@@ -22,6 +23,7 @@
         <td><%= link_to log.order.number, spree.edit_admin_order_path(log.order) %></td>
         <td><%= log.order_transaction&.transaction_id || "-" %></td>
         <td><%= log.status.capitalize %></td>
+        <td style="white-space: pre-wrap; width: 150px;"><%= log.error_message %></td>
         <td><%= log.created_at %></td>
         <td><%= log.updated_at %></td>
       </tr>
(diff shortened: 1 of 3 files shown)


app/views/spree/admin/transaction_sync_batches/show.html.erb
```diff
@@ -22,6 +23,7 @@
         <td><%= link_to log.order.number, spree.edit_admin_order_path(log.order) %></td>
         <td><%= log.order_transaction&.transaction_id || "-" %></td>
         <td><%= log.status.capitalize %></td>
+        <td style="white-space: pre-wrap; width: 150px;"><%= log.error_message %></td>
```
forkata, line 26: Thoughts on doing `.truncate(255)` or something along those lines here. Not sure if this could contain a stack trace, but if it does it'll be a very long and narrow column with this preset width.

forkata (approved):
Thanks @Noah-Silvera and @nvandoorn, left a comment on possibly truncating to a preset length, but otherwise this looks good to me.

forkata:
Also it looks like specs are running against 3.2 by default now so we may need to add `rails` to the `Gemfile` like we did in [this](https://github.com/SuperGoodSoft/solidus_taxjar/pull/193/commits/1ee569ee3861f658e7daeee26c50eb7cfde0064f#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR14) commit to get CI passing.

app/views/spree/admin/transaction_sync_batches/show.html.erb
forkata: We can ignore this for now since this won't include stack trace.

forkata (approved):

Noah-Silvera:
Just added a changelog commit and rebase on master!

Setting auto merge

PR #201, opened 2022-09-19
Support backfilling for specific date range

nvandoorn:
---

- Create some orders for a variety of date ranges
- Backfill the order for a specific date range
- Verify only the appropriate orders are backfilled

Merge Checklist
---

- [x] Run the manual tests
- [ ] Update the changelog

Screenshots
---


db/migrate/20220908181655_add_dates_to_transaction_sync_batch.rb
```diff
@@ -0,0 +1,6 @@
+class AddDatesToTransactionSyncBatch < ActiveRecord::Migration[6.1]
+  def change
+    add_column :solidus_taxjar_transaction_sync_batches, :start_date, :date
+    add_column :solidus_taxjar_transaction_sync_batches, :end_date, :date
```
forkata, line 4: I was originally thinking about us possibly storing these filters in a `serialized` column (similar to how we do preferences), so we don't have to migrate data when we add new filters. What are your thoughts on that approach, since it's unlikely we'll need to query on this data.

app/views/spree/admin/transaction_sync_batches/index.html.erb
```diff
@@ -4,8 +4,12 @@
   <%= "Transaction Sync Batches" %>
 <% end %>
 
-<%= form_with url: admin_transaction_sync_batches_path, method: :post, local: true do %>
-  <%= submit_tag "Backfill Transactions", class: "btn btn-primary" %>
+<%= form_with url: admin_transaction_sync_batches_path, method: :post, local: true do |f| %>
+  <%= f.label :start_date %>
+  <%= f.date_field :start_date %>
```
forkata, line 9: What do you think about providing a data/time input selector for extra granularity and storing this as a `DateTime`. I believe Solidus [has some partials](https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/backend/app/views/spree/admin/promotions/_form.html.erb#L47) for rendering date + time fields. You can see an example from promotions admin here
Screen Shot 2022-09-29 at 5 14 20 PM

forkata (commented):
This is looking great, thanks @Noah-Silvera and @nvandoorn!

benjaminwil (commented):
Reviewed!

Looks like this is still a draft and there's a spec failing. But I think it's looking great otherwise.

db/migrate/20220908181655_add_dates_to_transaction_sync_batch.rb
nvandoorn: In the interest of wrapping this ticket up I'm going to leave this for now. If we do end up updating this feature again this is a good suggestion.

spec/requests/spree/admin/transaction_sync_batches_request_spec.rb
```diff
@@ -16,16 +16,54 @@
 
     let!(:order) { create(:shipped_order) }
 
-    it "creates a batch with a log for the order" do
```
Noah-Silvera, line 19: I think we need to delete the cassette associated with this test name

Noah-Silvera (approved):

nvandoorn:
Latest push deletes the cassette @Noah-Silvera mentioned.

app/views/spree/admin/transaction_sync_batches/index.html.erb
nvandoorn: I'm not opposed to adding this feature in a new PR but I think using browser supplied date inputs is sufficient for now.

Noah-Silvera (approved):

Noah-Silvera:
Manually tested, setting auto merge!

PR #204, opened 2022-10-07
[186] Add page to display taxjar transactions

Noah-Silvera:
What is the goal of this PR?
---

Users should be able to easily view the transaction sync history for an order

---

1. Create a order with some taxjar transactions
    * [x] Ensure that you can click on the "TaxJar Sync History" tab and view the transactions
* [x] Smoke test viewing backfill logs

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

Screenshots
---

Screen Shot 2022-10-07 at 14 14 15PDT

### DIFF
--- a/app/overrides/spree/admin/orders_controller_override.rb
+++ b/app/overrides/spree/admin/orders_controller_override.rb
@@ -1,0 +1,11 @@
+module Spree
+  module Admin
+    module OrdersControllerOverride
+      def taxjar_transactions
+        load_order
+      end
+
+      Spree::Admin::OrdersController.prepend self
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
 - [#200](https://github.com/SuperGoodSoft/solidus_taxjar/pull/200) Render error message when backfilling transactions fails
 - [#203](https://github.com/SuperGoodSoft/solidus_taxjar/pull/203) Remove outdated backfill button
 - [#202](https://github.com/SuperGoodSoft/solidus_taxjar/pull/202) Add TaxJar transaction sync status to order show page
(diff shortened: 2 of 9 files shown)


benjaminwil (approved):
Other than the (trivial) spec failure that needs to be fixed. ✅

nvandoorn (approved):

Noah-Silvera:
Just a trivial fix to the tests!

app/overrides/spree/admin/orders_controller_override.rb
```diff
@@ -0,0 +1,11 @@
+module Spree
+  module Admin
+    module OrdersControllerOverride
```
forkata, line 3: Curious why we chose this path over adding a `TransactionSyncLogController` with an `index` action for this 🤔 In general I prefer we limit the amount of overriding and in this case it doesn't seem necessary.

Noah-Silvera: Mostly because it seemed like a nice touch to make it a tab on the orders page, instead of it's own page! You are right though comes at a cost to maintainability in the long run. Could go either way on this one!

forkata: > Mostly because it seemed like a nice touch to make it a tab on the orders page
I don't think this prevents us from using a separate controller. I don't think this action really belongs on the orders controller, and I think it would be more appropriate as a nested resource on orders.

I am also thinking we may eventually add a `#show` action if we need to display the full details of a log entry.

Noah-Silvera:
Just a small push to fix a spec!

app/overrides/spree/admin/orders_controller_override.rb
Noah-Silvera: Resolved IRL!

Noah-Silvera:
This push just fixed the `order_request_spec` so the tests pass on Solidus 3.0 and Solidus 3.1

PR #205, opened 2022-10-07
Use nexus regions for taxable address checks

### LINKED ISSUE #57: Check if merchant has nexus before calculating tax
Now that nexus regions are part of this extension, i want to propose updating the `taxable_address_check` to account for those. something like:

```ruby
    self.taxable_address_check = ->(address) {
      nexus_regions = ::SuperGood::SolidusTaxjar.api.nexus_regions
      nexus_regions.map(&:region_code).include?(address.state.abbr)
    }
```

This follows taxjars best practice as they have outlined in their API technical review:

```
2. API calls only for nexus states with the TaxJar API enabled

... 

Only call for a tax calculation in states where you have nexus
The /v2/nexus/regions endpoint can be used to create a dynamic filter to only call the TaxJar API for a tax calculation for states where you have nexus! This helps to control the number of API calls on a monthly basis!

You can accomplish this by creating a list based on the "region_code" from the response and comparing that to the state code that you would normally pass for the "to_state" in a /v2/taxes call.

Be sure to only pass the two-letter ISO code for the state when requesting a live tax calculation! If the state is passed as the full name, such as Arizona, the lookup will result in an error!
```

benjaminwil:
Closes #57.

What is the goal of this PR?
---

To meet a TaxJar certification requirement:

> Only send a request to v2/taxes when a merchant has nexus.

---

Not applicable. I've recorded TaxJar API requests that confirm that we're doing the right thing. Unless some of the United States change their state abbreviations, or a new U.S. state  appears and we need to upgrade the `carmen` gem, this should just work.

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog


benjaminwil:
The CI failures I am seeing are bewildering. I will look into them early next week. Please consider reviewing this PR regardless!

nvandoorn (approved):

benjaminwil:
Never mind. The VCR failures are related to this change. We'll be calling the TaxJar nexus endpoint much more frequently by default, now. I will need to re-record all of the effected cassettes.

benjaminwil:
@forkata and I reviewed this together in a meeting. As a result I think I need to revisit this:

- `taxable_address_check` might have additional business logic beyond nexus. It might be more straightforward to have a `has_nexus?` check that is separate from the taxable address check.
- We need to explicitly test for regions where nexus is by country and not by state.

Noah-Silvera:
> * We need to explicitly test for regions where nexus is by country and not by state.

@benjaminwil from what I'm reading on these guides we should skip this nexus check for any country that's not the US

> Here’s the most important takeaway for calculating sales tax: Only make US-based calculation requests where the merchant has nexus.

> * If an order is shipping to a merchant’s nexus state, make a request to /v2/taxes.
> * If an order is not shipping to a merchant’s nexus state, do not make a request. Ideally, fall back to your tax rate system.

https://developers.taxjar.com/integrations/sales-tax-calculations/

> * `taxable_address_check` might have additional business logic beyond nexus. It might be more straightforward to have a `has_nexus?` check that is separate from the taxable address check.

I think since the `taxable_address_check` is meant to be overridden and since this is a certification requirement, we should move into it's own non-overridable check like `has_nexus?` for sure!

benjaminwil:
@forkata and I have resolved all of the cassette-related failures and amended the final commit.

Noah-Silvera (approved):

PR #210, opened 2022-12-05
Add configuration settings documentation for v1.0

benjaminwil:
What is the goal of this PR?
---

Provide brief descriptions of available configuration settings.

---

Reviewers should read the documentation and check it against the implemented settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,20 +32,7 @@ introduced in Solidus v2.4. This maps better to how the TaxJar API itself works.
     bundle exec rails generate super_good:solidus_taxjar:install
     ```
 
-1. Also, configure your error handling:
-
-    ```ruby
-    # Put this in config/initializers/taxjar.rb
-
-    SuperGood::SolidusTaxjar.exception_handler = ->(e) {
-      # Report exceptions in here. For example, if you were using the Sentry's
-      # raven-ruby gem to report errors, you might do this:
-      Raven.capture_exception(e)
-    }
-    ```
-
-    For more information about configuring the extension, see
-    [Configuration](#configuration).
+1. Also, configure your own [exception handler](#exception-handling).
 
 1. Finally, make sure that the `TAXJAR_API_KEY` environment variable is set to
   your TaxJar API key.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,10 +106,111 @@ SuperGood::SolidusTaxjar.tap do |config|
...
+	sent.
+
+	Note that all the essential `Spree::Order`, `Spree::LineItem`, and
+	`Spree::Address` attributes are already sent by default.
+
+	Default value: `{}`
+
+- `discount_calculator`: A discount calculator class. A safe default
+  calculator is provided, but may not be suitable for
+  stores. Check with your accountant how discounts need
```
Noah-Silvera, line 136: I think the main concern here is that our discount calculator doesn't handle order level adjustments. 

Instead of asking users to talk to their accountants, maybe we should phrase it like this!

```
A discount calculator class. A default
calculator is provided, but may not be suitable for
stores. TaxJar requires discounts to be specified per
line item, so a custom calculator will be required to
support order level discounts and other promotions.
```

README.md
```diff
@@ -112,10 +106,111 @@ SuperGood::SolidusTaxjar.tap do |config|
...
+  to be calculated.
+
+  Default value: `::SuperGood::SolidusTaxjar::DiscountCalculator`
+
+- `exception_handler`: A function that returns your own exception
+  handler. See [Exception handling](#exception-handling).
+
+  Default value: (See the source code.)
+
+- `job_queue`: A string or symbole that matches the name of a
```
Noah-Silvera, line 146: symbol**é**

image

README.md
```diff
@@ -112,10 +106,111 @@ SuperGood::SolidusTaxjar.tap do |config|
...
+  for accurate calculations.
+
+  Default value: (See the source code.)
+
+- `shipping_tax_label_maker`: A string to label shipment taxes in TaxJar.
+
+  Default value: "Sales Tax"
+
+- `taxable_address_check`: A function that returns a boolean after checking
+  whether an address should be considered taxable.  By default this function
```
Noah-Silvera, line 172: After [this PR](https://github.com/SuperGoodSoft/solidus_taxjar/pull/205), the default will be true unless you are in the united states and the address is **not** in a nexus region

Noah-Silvera (commented):
Thanks for kicking the documentation ticket off Benjamin!!

README.md
forkata:
```
A default calculator is provided, but may not be suitable for stores.
```

I think this statement generalizes a bit and for most stores might not be true. We shouldn't assume that the default calculator is not suitable for most stores, since promotions which rely on order level adjustments are _known_ to not work well with the built-in tax system as well so they are not widely used anymore.

I think as an extension whose main purpose is to provide tax calculations, we should probably phrase this in a way that mentions the exception to the rule, rather than start off with "this might not work for you".

```
A discount calculator class. A default calculator is provided which handles promotions which rely on line item and shipment adjustments. Since TaxJar requires discounts to be specified per line item, you will need to provide a custom version of this calculator in order to support promotions which create order level adjustments. Adjustments to an order made through the Solidus admin would also need to be handled through a custom implementation of this class.
```

README.md
benjaminwil: Okay. I will leave as is, and if this gets merged first I can add a commit to that PR.

benjaminwil:
@forkata @Noah-Silvera 

Thanks for the review. I've addressed your feedback and added some more documentation:

- I've added other 1.0 features to the **Features** section.
- I added a new section, **Order-level Adjustments are Unsupported** to explain to developers what they need to do to support order-level adjustments should they need to.

forkata (approved):

README.md
benjaminwil: @forkata and I just double-checked. We've refactored the linked PR and we no longer change the `taxable_address_check`'s default return value.

PR #215, opened 2023-01-28
Update CHANGELOG to "Keep a Changelog" format

forkata:
What is the goal of this PR?
---
This updates the CHANGELOG file to use the format suggested by https://keepachangelog.com and breaks up the changes which will be included in the next release into a few separate categories. It also includes a section at the top titled `Unreleased` where we will add all future changes. That will hopefully help people who use the extension through Rubygems see what will be coming in the next official release.

I am hoping this format help users of the extension easily identify the changes they need to be aware of when upgrading to a new release and also provide us with a framework for how to consistently identify and group related changes.

I looked at the format [solidusio/solidus](https://github.com/solidusio/solidus/blob/master/CHANGELOG.md) uses for their CHANGELOG file, but did not find that to be as helpful as this. That file just has a big section of `Major changes` under each release, which doesn't really help quickly identify what you need to be aware of. I personally find breaking out the additive changes from the removals and changes in behaviour to be useful when upgrading gem versions.

I am proposing this format, but would love to hear if others have suggestions for a different one. 

---

N/A

Merge Checklist
---
- [ ] ~Finalize the changes that will be included in the 1.0.0 release~
- [ ] ~Update the release date for the 1.0.0 release~

**Update:** We are going to merge this before we've finalized the 1.0 release so this no longer applies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,14 @@
 # Changelog
 
-## master
+All notable changes to this project will be documented in this file.
 
-- [#250](https://github.com/SuperGoodSoft/solidus_taxjar/pull/250) Improve error handling for `ReportTransactionJob`
-- [#249](https://github.com/SuperGoodSoft/solidus_taxjar/pull/249) Don't depend on `webdrivers` gem
-- [#244](https://github.com/SuperGoodSoft/solidus_taxjar/pull/244) Let admins manually retry syncing TaxJar transactions
-- [#193](https://github.com/SuperGoodSoft/solidus_taxjar/pull/193) Bump version requirement for `solidus_support` to ">= 0.9.0", and as a result drop support for Rails 5.2.
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+### Added
+
 - [#190](https://github.com/SuperGoodSoft/solidus_taxjar/pull/190) Add transaction sync batch show page
 - [#188](https://github.com/SuperGoodSoft/solidus_taxjar/pull/188) Show transaction sync batches in user interface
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -38,25 +41,46 @@
...
 - [#204](https://github.com/SuperGoodSoft/solidus_taxjar/pull/204) Add page to render an orders TaxJar transactions
 - [#201](https://github.com/SuperGoodSoft/solidus_taxjar/pull/201) Allow transaction backfills to be filtered by date range
+
+### Changed
+
+- [#95](https://github.com/SuperGoodSoft/solidus_taxjar/pull/95) Only require "state" for Canadian and US addresses
+- [#120](https://github.com/SuperGoodSoft/solidus_taxjar/pull/120) Change default `SOLIDUS_BRANCH` and `RAILS_VERSION`
+- [#137](https://github.com/SuperGoodSoft/solidus_taxjar/pull/137) Run tests against the most up to date versions of Solidus
+- [#181](https://github.com/SuperGoodSoft/solidus_taxjar/pull/181) Take all non-tax adjustment types into account when calculating a line item's discount
+- [#182](https://github.com/SuperGoodSoft/solidus_taxjar/pull/182) Automatically create default Tax Rate
```
benjaminwil, line 63: Only three of these items effect end users:

- Only require "state" for Canadian and US addresses
- Take all non-tax adjustment types into account when calculating a line item's discount
- Automatically create default Tax Rate

What do you think about separating this list into "Changes for end users" and "Changes for extension developers" or something like that?

forkata: That makes sense. We could break out a separate section for extension developers :+1:

benjaminwil (approved):

jarednorman (approved):

forkata:
@Noah-Silvera I rebased this and updated it for the changes we've merged since I opened it. I think we should merge this now so it's ready for release, we just need to change the `Unreleased` to `1.0 (datestamp)` when we're ready to publish that release.

benjaminwil (approved):

Issue #212, opened 2023-01-25
uninitialized constant Deface


diegomichel:
In case someone runs into the same issue:

```
.rvm/gems/ruby-3.0.2@project/gems/super_good-solidus_taxjar-0.18.2/app/overrides/spree/admin/shared/_configuration_menu.rb:3:in `<main>': uninitialized constant Deface (NameError)
```

workaround: add this to your Gemfile project:

```
gem 'deface'
```

forkata:
Good call on this, we should add this to the gemspec of this extension! It looks like we're getting it in the dummy app against which we run tests from `solidus_auth_device` and that's why we never realized we need to explicitly depend on that!

diegomichel:
@forkata I tried to add it to the `gemspec`, then I had to require it manually on the view, but after that, I ran into other issues

```
.../bundler/gems/solidus_taxjar-8e81b0375cb1/app/overrides/spree/admin/shared/_configuration_menu.rb to define constant Spree::Admin::Shared::ConfigurationMenu, but didn't (Zeitwerk::NameError)   
```

https://github.com/SuperGoodSoft/solidus_taxjar/compare/master...diegomichel:solidus_taxjar:master

For now the workaround will do :)

forkata:
This is now fixed! Thanks @benjaminwil!

PR #216, opened 2023-01-30
Add `deface` as dependency

benjaminwil:
What is the goal of this PR?
---

Because our test application includes `solidus_auth_devise`, which depends on `deface`, we didn't realize that we needed to explicitly depend on `deface`. We wrongly thought it would be available to all Solidus stores.

This commit should resolve issue #212.

---

I would like to manually ensure the admin interface is working as intended to ensure we aren't creating problems for end users, like reported in [this comment](https://github.com/SuperGoodSoft/solidus_taxjar/issues/212#issuecomment-1404261063).

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

### DIFF
--- a/lib/super_good/solidus_taxjar.rb
+++ b/lib/super_good/solidus_taxjar.rb
@@ -1,5 +1,6 @@
 require "solidus_core"
 require "solidus_support"
+require "deface"
 require "taxjar"
 require "super_good/solidus_taxjar/overrides/request_override"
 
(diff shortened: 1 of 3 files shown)


lib/super_good/solidus_taxjar.rb
```diff
@@ -1,3 +1,4 @@
+require "deface"
```
forkata, line 1: We may need to move this after `solidus_core` 🤔 It looks like Rails is required but not loaded yet at this point in time.

benjaminwil: Ah, thank you. Resolved.

Noah-Silvera (approved):

forkata (approved):

forkata:
Rebased to fix a conflict in the CHANGELOG

PR #217, opened 2023-02-06
Handle updates to orders that have not been reported

nvandoorn:
What is the goal of this PR?
---

To gracefully handle updates to orders that were never reported at time of purchase.

---

1. Ensure reporting is disabled
1. Complete an order
1. Enable reporting
1. Issue a refund (or some other change that fires `order_recalculated`)
1. Verify the order is not reported
1. Verify there is a sync log with an error

Merge Checklist
---

- [x] Run the manual tests
- [ ] Update the changelog

### DIFF
--- a/app/controllers/spree/admin/taxjar_settings_controller.rb
+++ b/app/controllers/spree/admin/taxjar_settings_controller.rb
@@ -41,7 +41,10 @@ module Spree
       end
 
       def configuration_params
-        params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled)
+        params_hash = params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled, :preferred_reporting_enabled_at_integer)
+        {
+          preferred_reporting_enabled_at_integer: params_hash[:preferred_reporting_enabled] == "1" ? DateTime.now.to_i : nil
+        }
       end
 
       def tax_category_params
--- a/app/models/super_good/solidus_taxjar/configuration.rb
+++ b/app/models/super_good/solidus_taxjar/configuration.rb
@@ -9,7 +9,8 @@ module SuperGood
       preference :reporting_enabled_at_integer, :integer, default: nil
 
       def preferred_reporting_enabled
-        preferred_reporting_enabled_at_integer.present?
+        preferred_reporting_enabled_at_integer.present? &&
+          DateTime.current.after?(preferred_reporting_enabled_at)
       end
 
       def preferred_reporting_enabled_at
--- a/app/models/super_good/solidus_taxjar/configuration.rb
+++ b/app/models/super_good/solidus_taxjar/configuration.rb
@@ -6,7 +6,15 @@ module SuperGood
       include ::Spree::Preferences::Persistable
 
       self.table_name = 'solidus_taxjar_configuration'
-      preference :reporting_enabled, :boolean, default: false
+      preference :reporting_enabled_at_integer, :integer, default: nil
+
+      def preferred_reporting_enabled
+        preferred_reporting_enabled_at_integer.present?
+      end
+
+      def preferred_reporting_enabled_at
+        Time.at(SuperGood::SolidusTaxjar.configuration.preferred_reporting_enabled_at_integer).to_datetime
+      end
 
       class << self
         def default
--- a/app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
+++ b/app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
@@ -21,6 +21,7 @@ module SuperGood
 
           if completed_before_reporting_enabled?(order)
             order.taxjar_transaction_sync_logs.create!(status: :error, error_message: ORDER_TOO_OLD_MESSAGE)
+            SuperGood::SolidusTaxjar.exception_handler.call(RuntimeError.new(ORDER_TOO_OLD_MESSAGE))
             return
           end
 
@@ -36,6 +37,7 @@ module SuperGood
 
           if completed_before_reporting_enabled?(order)
             order.taxjar_transaction_sync_logs.create!(status: :error, error_message: ORDER_TOO_OLD_MESSAGE)
+            SuperGood::SolidusTaxjar.exception_handler.call(RuntimeError.new(ORDER_TOO_OLD_MESSAGE))
             return
           end
 
--- a/app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
+++ b/app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
@@ -15,16 +15,23 @@ module SuperGood
         def report_transaction(event)
           return unless SuperGood::SolidusTaxjar.configuration.preferred_reporting_enabled
 
-          SuperGood::SolidusTaxjar::ReportTransactionJob.perform_later(event.payload[:shipment].order)
+          order = event.payload[:shipment].order
+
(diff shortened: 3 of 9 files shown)


app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
```diff
@@ -19,6 +19,7 @@ def report_transaction(event)
 
           unless order.completed_at && SuperGood::SolidusTaxjar.configuration.preferred_reporting_enabled_at > order.completed_at
             order.taxjar_transaction_sync_logs.create!(status: :error, error_message: ORDER_TOO_OLD_MESSAGE)
+            SuperGood::SolidusTaxjar.exception_handler.call(RuntimeError.new(ORDER_TOO_OLD_MESSAGE))
```
forkata, line 22: I am a bit torn on this, if it's an expected outcome that we don't report orders that are created prior to the backfill period start, is it really an exception?

forkata (commented):
Thanks for tackling this @Noah-Silvera and @nvandoorn 🎉 I left a couple of nits, but otherwise great work!

app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
```diff
@@ -13,18 +13,20 @@ module ReportingSubscriber
...
+          return unless order.completed_at && SuperGood::SolidusTaxjar.configuration.preferred_reporting_enabled_at > order.completed_at
+
+          SuperGood::SolidusTaxjar::ReportTransactionJob.perform_later(order)
         end
 
         def replace_transaction(event)
           order = event.payload[:order]
 
-          return unless SuperGood::SolidusTaxjar.configuration.preferred_reporting_enabled
+          return unless order.completed_at && SuperGood::SolidusTaxjar.configuration.preferred_reporting_enabled_at > order.completed_at
```
benjaminwil: I don't feel strongly about this, but to make it easier to reason about these guard statements: it _could_ be nice to move them into a private method that describes why we return early:

```suggestion
          return unless reportable?(order)
```

Now, we have a concept of "reportability" with two requirements. And if an end user needed to customize their own reportability requirements they wouldn't have to override the subscribers.

app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
benjaminwil: I don't feel strongly about this. A possible alternative: what would be the drawbacks of emitting an INFO level log line instead?

benjaminwil (approved):

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -41,7 +41,13 @@ def cached_api
       end
 
       def configuration_params
-        params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled)
+        params_hash = params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled, :preferred_reporting_enabled_at_integer)
+        if params_hash.delete(:preferred_reporting_enabled)
+          params_hash[:preferred_reporting_enabled_at_integer] = DateTime.now.to_i
```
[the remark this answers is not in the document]
Noah-Silvera: The timezone actually doesn't matter here as `to_i` converts it to a fixed point in time, as an integer.

`Time.zone.now.to_datetime.to_i` and `DateTime.now.to_i` return the same thing regardless of timezones!

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -41,7 +41,13 @@ def cached_api
...
+        else
+          params_hash[:preferred_reporting_enabled_at_integer] = nil
+        end
```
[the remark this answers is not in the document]
Noah-Silvera: Done!

spec/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber_spec.rb
```diff
@@ -49,10 +48,9 @@ def with_events_disabled(&block)
         }
 
         it "does nothing" do
-          expect(reporting)
-            .not_to receive(:refund_and_create_new_transaction)
-
           subject
+
+          assert_no_enqueued_jobs
```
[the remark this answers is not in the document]
Noah-Silvera: Good point! I updated the commit message

app/controllers/spree/admin/taxjar_settings_controller.rb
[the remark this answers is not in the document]
benjaminwil: oh!

app/subscribers/super_good/solidus_taxjar/spree/reporting_subscriber.rb
forkata: Discussed IRL and the consensus here the more visibility the better.

The reason an INFO log message is not ideal is that it could possibly go nowhere, so we are leveraging the user configured exception handler instead.

app/models/super_good/solidus_taxjar/configuration.rb
```diff
@@ -6,7 +6,15 @@ class Configuration < ::Spree::Base
...
       self.table_name = 'solidus_taxjar_configuration'
-      preference :reporting_enabled, :boolean, default: false
+      preference :reporting_enabled_at_integer, :integer, default: nil
+
+      def preferred_reporting_enabled
+        preferred_reporting_enabled_at_integer.present?
+      end
+
+      def preferred_reporting_enabled_at
+        Time.at(SuperGood::SolidusTaxjar.configuration.preferred_reporting_enabled_at_integer).to_datetime
```
forkata, line 16: We should add some documentation to the configuration in the Readme for this 🙏🏼 

From a UX perspective, using an Epoch based time here seems a bit weird to me, did you consider storing this as a string in a standard datetime format (ISO 8601), i.e. "2022-09-27 18:00:00.000". We would probably want this to be set as UTC time and converted to the Rails configured timezone.

app/controllers/spree/admin/taxjar_settings_controller.rb
```diff
@@ -41,7 +41,10 @@ def cached_api
       end
 
       def configuration_params
-        params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled)
+        params_hash = params.require(:super_good_solidus_taxjar_configuration).permit(:preferred_reporting_enabled, :preferred_reporting_enabled_at_integer)
+        {
+          preferred_reporting_enabled_at_integer: params_hash[:preferred_reporting_enabled] ? DateTime.now.to_i : nil
```
forkata, line 46: I am a bit confused about this. Is this the date we will be using for back-filling transactions? If so I don't think we want to rely on someone hitting the enable button at the time when they want that to start?

If I recall the backfill dashboard has a date picker, should we be basing it on that instead or using this date there 😕

forkata (commented):
Thanks for addressing the comments I left, I still have a few questions about how we are handling the reporting enabled date and how that works with regard to back-filling transactions.

Noah-Silvera:
Cleaned up the commits, updated the changelog, this is ready to go!

benjaminwil (approved):

forkata:
@nvandoorn and I did a deep dive on the spec failures and it looks like something in Solidus 3.2 changed which results in our spec setup enqueueing a `ReplaceTransactionJob`. It's possible that we just need to reset the job queue between specs so there is no leakage, but for now we've fixed the failures by just ensuring that calling the subject doesn't change the count.

PR #227, opened 2023-04-14
Fix reporting for orders with refunds

nvandoorn:
What is the goal of this PR?
---

Prior to this change, refunds would not trigger the replace transaction job. This is because we compared `order.total` which does not include refunds/reimbursements. We fix this by calling `order.payment_total` instead as it reflects what the customer actually paid. Finally, we have to make the same change in the reporting code.

---

1. Ensure reporting is enabled
1. Complete and ship a new order
1. Refund one or more of the line items
1. Verify the order has been re-reported to TaxJar with the new amount.

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

### DIFF
--- a/spec/super_good/solidus_taxjar/api_params_spec.rb
+++ b/spec/super_good/solidus_taxjar/api_params_spec.rb
@@ -12,9 +12,11 @@ RSpec.describe SuperGood::SolidusTaxjar::ApiParams do
       total: order_total,
       shipments: [shipment],
       user_id: 12345,
-      completed_at: DateTime.new(2018, 3, 6, 12, 10, 33))
+      completed_at: DateTime.new(2018, 3, 6, 12, 10, 33),
+      payment_total: payment_total)
   end
   let(:order_total) { BigDecimal("123.45") }
+  let(:payment_total) { BigDecimal("110.45") }
 
   let(:store) do
     create(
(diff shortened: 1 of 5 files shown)


benjaminwil:
If we rebase against master the CI run should go much better. ✨

forkata (approved):
This code change looks good, but I think we may want to get confirmation that this is the correct way for us to handle refunds/reimbursements. There is some information on this article and the associated blog post linked at the end, but it's more around the filing process - https://support.taxjar.com/article/250-how-are-refunds-displayed-in-my-reports

spec/super_good/solidus_taxjar/api_params_spec.rb
```diff
@@ -12,9 +12,11 @@
       total: order_total,
       shipments: [shipment],
       user_id: 12345,
-      completed_at: DateTime.new(2018, 3, 6, 12, 10, 33))
+      completed_at: DateTime.new(2018, 3, 6, 12, 10, 33),
+      payment_total: payment_total)
   end
   let(:order_total) { BigDecimal("123.45") }
```
benjaminwil, line 18: Should we remove the `order_total` assignment and the `#total` parameter?

benjaminwil (approved):
I think that this work is required and is more accurate than what was committed before.

Based on the article Chris linked to, it'd be nice to investigate what we can do about reporting periods. But it looks like the API doesn't give us a clean way to post refund transactions to a specific reporting period, so I'm currently at a loss.

AlistairNorman (approved):

forkata:
> But it looks like the API doesn't give us a clean way to post refund transactions to a specific reporting period, so I'm currently at a loss.

The way I read the blog post is that TaxJar handles that for you and now they even know to not submit negative reports but roll them up to the next reporting period if say you only have a refund for a given state.

PR #232, opened 2023-05-19
Update how we load factories to be more compatible with Solidus

### LINKED ISSUE #230: Fix factory override
Our specs are failing on Solidus 3.1 and newer due to a change in how factories are loaded and as a result how we need to do overrides to factories from core.

More details on the upstream change here -  https://github.com/solidusio/solidus/discussions/4014

### What

We have an override for the `Spree::Address` factory that needs to be updated, see failing specs here - https://app.circleci.com/pipelines/github/SuperGoodSoft/solidus_taxjar/758/workflows/c11e6090-3ac2-43c7-b491-3fababba73f1

forkata:
What is the goal of this PR?
---
This is a temporary fix until our PR to `solidus_dev_support` is merged
that allows us to have deeper namespacing of the extension folder
structure.

https://github.com/solidusio/solidus_dev_support/pull/207

This change also updates how we run the CI jobs to separate the current / older versions into separate jobs. With the new Solidus v4.0 release, we still have some work to get the extension passing agains that version as well as `main`. Splitting the jobs allows us to also run each one with a different version of Ruby, which is now required as the `solidus_sample` repo requires Ruby `>= 3.0` for Solidus v4 and `main`.

**NOTE:** There is some overlap in the test runs as `run-specs-solidus-current` will run tests on `v3.2`, `v3.3`, `v3.4` and `v4.0` (on Ruby 3) and `run-specs-solidus-older` will run specs on `v3.2`, `v3.3` and `v3.4` (on Ruby 2.7). The last one is what I changed the Required check to be on, since we’re not passing on 4.0 yet.

Fixes #230.

---
* Make sure specs pass against _older_ Solidus versions ~including `main`~

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -24,8 +24,17 @@ Capybara.register_driver :solidus_chrome_headless do |app|
   Capybara::Selenium::Driver.new(app, browser: :chrome, options_key => chrome_options)
 end
 
-factories = Dir["#{::SuperGoodSolidusTaxjar::Engine.root}/lib/super_good/solidus_taxjar/testing_support/factories/**/*_factory.rb"].sort
-factories.each { |f| require f }
+# This was pulled from `solidus_dev_support` and the glob was updated to allow
+# for a deeper namespacing of our factories folder
+# `lib/solidus_dev_support/testing_support/factories.rb`.
+paths = ::SuperGoodSolidusTaxjar::Engine.root.glob('lib/**/testing_support/factories')
+
+FactoryBot.definition_file_paths = [
+  Spree::TestingSupport::FactoryBot.definition_file_paths,
+  paths,
+].flatten
+
+FactoryBot.reload
 
 # Requires supporting ruby files with custom matchers and macros, etc,
 # in spec/support/ and its subdirectories.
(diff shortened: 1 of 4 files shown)


nvandoorn (approved):

spec/spec_helper.rb
```diff
@@ -24,8 +24,17 @@
   Capybara::Selenium::Driver.new(app, browser: :chrome, options_key => chrome_options)
 end
 
-factories = Dir["#{::SuperGoodSolidusTaxjar::Engine.root}/lib/super_good/solidus_taxjar/testing_support/factories/**/*_factory.rb"].sort
-factories.each { |f| require f }
+# This was pulled from `solidus_dev_support` and the glob was updated to allow
+# for a deeper namespacing of our factories folder
+# `lib/solidus_dev_support/testing_support/factories.rb`.
+paths = ::SuperGoodSolidusTaxjar::Engine.root.glob('lib/**/testing_support/factories')
```
benjaminwil, line 30: 🔥

benjaminwil (approved):

forkata:
It looks like we are now failing specs on Solidus 4.0, we can make that check optional, but I am also unsure where to rename the `Required` check from the old one?

This is the failure we need to fix for tests to pass on `main` and `v4.0`
```
Failure/Error: include ::Spree::Event::Subscriber

NameError:
  uninitialized constant Spree::Event
```

nvandoorn (commented):
@forkata Last commit to split things up looks good!

PR #235, opened 2023-06-21
Add refund feature spec

nvandoorn:
What is the goal of this PR?
---

Prior to this commit, testing the refund feature manually was cumbersome as we developed fixes for this feature.

### DIFF
--- a/spec/features/spree/admin/refund_spec.rb
+++ b/spec/features/spree/admin/refund_spec.rb
@@ -1,0 +1,52 @@
+require 'spec_helper'
+
+RSpec.feature 'Refunding an order', js: true do
+  stub_authorization!
+
+  let(:order) { create(:shipped_order, line_items_count: 3) }
+
+  before do
+    country = create(:country, states_required: true)
+    create(:product, name: "RoR Mug")
+    create(:store, default: true)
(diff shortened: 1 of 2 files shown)


AlistairNorman (approved):

benjaminwil (approved):

spec/features/spree/admin/refund_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    click_on "Create reimbursement"
+    select "Original payment", from: "reimbursement[return_items_attributes][0][override_reimbursement_type_id]"
+    click_on "Update"
+
+    perform_enqueued_jobs do
+      click_on "Reimburse"
+    end
+
+    expect(find(".reimbursement-refund-amount")).to have_content("$10.89")
```
forkata, line 50: Are we planning on adding any assertions for the transaction being reported to TaxJar or is that going to be done in a different PR?

forkata (approved):

spec/features/spree/admin/refund_spec.rb
benjaminwil: It would be nice to add assertions that verify that the reporting was successful if that is exposed in the admin interface somewhere.

PR #236, opened 2023-06-26
[221] Don't send cancelled or returned line items

Noah-Silvera:
Prior to this change, each request to TaxJar sent all line items. We
should exclude line items that have been returned or cancelled so the
total of the line items still adds up to the order's payment total.

### DIFF
--- a/lib/super_good/solidus_taxjar/api_params.rb
+++ b/lib/super_good/solidus_taxjar/api_params.rb
@@ -146,7 +146,8 @@ module SuperGood
           # The API appears to error when sent line items with no quantity...
           # but why would you do that anyway.
           line_items.reject do |line_item|
-            line_item.quantity.zero?
+            line_item.quantity.zero? ||
+              !line_item.inventory_units.where.not(state: ['returned', 'canceled']).exists?
           end
         end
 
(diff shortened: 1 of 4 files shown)


lib/super_good/solidus_taxjar/api_params.rb
```diff
@@ -146,7 +146,8 @@ def valid_line_items(line_items)
           # The API appears to error when sent line items with no quantity...
           # but why would you do that anyway.
           line_items.reject do |line_item|
-            line_item.quantity.zero?
+            line_item.quantity.zero? ||
+              !line_item.inventory_units.where.not(state: 'returned').exists?
```
forkata: Could we write this as
```
line_item.inventory_units.returned.any?
```

lib/super_good/solidus_taxjar/api_params.rb
```diff
@@ -116,7 +116,7 @@ def line_items_params(line_items)
             line_items: valid_line_items(line_items).map do |line_item|
               {
                 id: line_item.id,
-                quantity: line_item.quantity,
+                quantity: line_item.inventory_units.where.not(state: ['returned', 'canceled']).count,
```
forkata, line 119: Since we're doing this a few times, should we consider pulling it into a helper so we don't have to repeat it?

benjaminwil (approved):
I'll add: I like both of Chris's comments.

nvandoorn:
Latest revision rebases this on-top of master and moves the repeated query into a private method.

forkata (approved):
Thanks for the clean-up here @nvandoorn 🚀

benjaminwil (approved):

PR #240, opened 2023-07-14
Hard-code order number in refund feature spec

nvandoorn:
Prior to this change, this spec may fail because the order number
differs from the order number in the cassette. To solve this, we
hard-code the value from the cassette into the test.

Note that re-recording this cassette may require incrementing the order
number.

What is the goal of this PR?
---

---

1. Do a thing
    * [ ] Assert a result

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

Screenshots
---

### DIFF
--- a/spec/features/spree/admin/refund_spec.rb
+++ b/spec/features/spree/admin/refund_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 RSpec.feature 'Refunding an order', js: true do
   stub_authorization!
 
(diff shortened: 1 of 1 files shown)


forkata (approved):

AlistairNorman (approved):

nvandoorn:
Drafting this until we remove the `pending` from the refund feature spec.

benjaminwil:
@nvandoorn Is this closed forever? Did we determine this isn't an issue?

PR #241, opened 2023-07-14
Update refund feature spec cassette

nvandoorn:
What is the goal of this PR?
---

---

1. Do a thing
    * [ ] Assert a result

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

Screenshots
---


nvandoorn:
@benjaminwil and I worked on debugging some incorrect amounts in the cassettes for the refund feature spec. We figured the cassette may need to be re-recorded, but this does not seem to be the case. Instead it appears to be a race condition in the refund feature spec. Closing this to open a pull request to fix the race condition.

PR #246, opened 2023-09-26
ISSUE-223 | Report instead of trying to replace transactions that failed to sync on recalculate

### LINKED ISSUE #223: Orders will not be re-synced to TaxJar if first sync fails
If an order's first sync to TaxJar fails, we may end up calling `replace_transaction` instead of `report_transaction` which results in a no-op. We need a mechanism to retry syncing to TaxJar when the first sync fails.

forkata:
What is the goal of this PR?
---

This change aims to fix the issue with replacing a transaction on `recalculate`
when it has previously failed to be reported to TaxJar.

**Update:** This change ended up being a bigger refactor and resulted in us consolidating the two Solidus events we were sync'ing orders to TaxJar on into a single event - the `order_recalculated` one.

Fixes #223 

---

1. Do a thing
    * [ ] Assert a result

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

Screenshots
---


Noah-Silvera:
Here are the failure cases we need to handle

## Case 1: Initial reporting fails
* The initial reporting fails (potentially multiple times)

## Case 2: Replacing a transaction totally fails
* The initial reporting succeeds
* The call to for the REFUND transaction fails

## Case 3: Replacing a transaction partially fails
* The initial reporting succeeds
* The call for the REFUND transaction succeeds
* The call for the new transaction **fails**

## Case 4: Repeated failures
 * The initial reporting succeeds
* The call for the REFUND transaction suceeds
* The call for the new transaction fails
* The call for the new transaction succeeds
	* Another replacement happens
* The call for the REFUND transaction succeeds
* The call for the new transaction **fails**

forkata:
After some discussion with @Noah-Silvera, we are going to try and unify the `report` and `replace` event handlers into a single event in order to reduce some of the complexity in reporting transactions.

lib/super_good/solidus_taxjar/reportable.rb
```diff
@@ -46,14 +46,19 @@ def order_reportable?(order)
...
       def transaction_replaceable?(order)
-        order.taxjar_order_transactions.present? &&
-          order.payment_state == "paid" &&
+        return unless order.payment_state == "paid"
+
+        if order.taxjar_order_transactions.present?
           amount_changed?(order)
+        else
+          true
+        end
```
benjaminwil: Looks like you came to the same conclusion based on your recently comments on the PR. But I think we should not merge this change as-is. With this change, reporting and replacing isn't meaningfully different, so we should just merge them into a single function.

benjaminwil: Just wanted to highlight this change for further discussion if necessary.

forkata:
For some reason psych v4 is getting pulled in here. We may want to retry the specs on master to see if this isn't an upstream change.

forkata:
Looks like the specs are failing against `main` as well, so we can fix there https://app.circleci.com/pipelines/github/SuperGoodSoft/solidus_taxjar/908/workflows/0129bc98-5f88-4555-8a94-7beda08b5af0/jobs/1527

Noah-Silvera:
Cleaned up the commit history here! Going to try running some manual tests

Noah-Silvera:
> Here are the failure cases we need to handle
> 
> ## Case 1: Initial reporting fails
> * The initial reporting **fails** (potentially multiple times)
> **SUCCESS**
> ## Case 2: Replacing a transaction totally fails
> * The initial reporting **succeeds**
> * The call to for the REFUND transaction **fails**
> **SUCCESS**
> ## Case 3: Replacing a transaction partially fails
> * The initial reporting **succeeds**
> * The call for the REFUND transaction **succeeds**
> * The call for the new transaction **fails**
> **FAILURE**

Screen Shot 2023-12-15 at 11 11 23 AM


When the call for the new transaction fails, the overall transaction sync is still shown as "succeeded"

Noah-Silvera:
I ran through all the tests and was able to successfully handle the error cases!

Noah-Silvera:
This push should fix the specs!

forkata:
✅ Can't approve, but this looks good to me! I think we should write a changelog entry explaining the changes here, but otherwise I think this is ready for some testing 🚀

Noah-Silvera:
> ✅ Can't approve, but this looks good to me! I think we should write a changelog entry explaining the changes here, but otherwise I think this is ready for some testing 🚀

Sweet!! @forkata I already ran the tests! All the manual test cases I listed above have been run :)

Issue #248, opened 2023-10-23
BackfillTransactionSyncBatchJob errors out and doesn't finish if a transaction job raises an error thats not a Taxjar::Error


AdnanTheExcellent:
As the title, says, if any ReportTransactionJob.perform is run and there's an error thats not a `Taxjar::Error`, the error will bubble up and end the transaction sync batch job. 

I was running a batch of ~ 1200 orders and the batch kept stopping in the single digits. Once i got rid of the problem order(s), the batch ran all the way through. 

![image](https://github.com/SuperGoodSoft/solidus_taxjar/assets/943069/61c966b9-859d-4f79-9cdc-b409dbcdb13b)

Solution: rescue `StandardError` in addition to the `Taxjar::Error`. This won't prevent timeouts from stopping a batch from running, but it will prevent most instances where a stray error might fail the batch. 

Something like this in in ReportTransactionJob.perform:

```ruby
  begin
    order_transaction = SuperGood::SolidusTaxjar.reporting.show_or_create_transaction(order)
    transaction_sync_log.update!(order_transaction: order_transaction, status: :success)
  rescue Taxjar::Error => exception
    transaction_sync_log.update!(status: :error, error_message: exception.message)
  rescue StandardError => exception
    Rails.logger.error exception.backtrace.join("\n").to_s
    transaction_sync_log.update!(status: :error, error_message: "An error occurred")
  end
```

PR #250, opened 2023-11-21
Improve error handling for `ReportTransactionJob`

benjaminwil:
What is the goal of this PR?
---

This pull request resolves a user reported error, #248.

Merge Checklist
---

- [x] Update the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master
 
+- [#250](https://github.com/SuperGoodSoft/solidus_taxjar/pull/250) Improve error handling for `ReportTransactionJob`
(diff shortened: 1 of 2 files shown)


forkata (approved):
LGTM, should we add a spec for this change?

benjaminwil:
There is an existing spec for raising an error from the job. It didn't seem valuable to me to make a second one with a different error class.

forkata:
> There is an existing spec for raising an error from the job. It didn't seem valuable to me to make a second one with a different error class.

We could change the type of error raised I guess 🤔

benjaminwil:
Okay, I have been convinced. I will add an additional test for `StandardError` rescuing specifically.

forkata:
Rebased to resolve conflicts.

PR #253, opened 2024-01-16
Resolve reimbursement tax calculation discrepancies

benjaminwil:
What is the goal of this PR?
---

While we were manually testing the extension for the 1.0 release, we discovered that going through the stock Solidus RMA -> customer return -> reimbursement flow would result in reimbursements being generated with the wrong amount of tax.

After the manual testing session, I discovered that the refund feature test was being skipped. I refactored it so we can more easily test two scenarios:

  1. Removing an entire line item results in the correct reimbursement amount being generated.
  2. Removing part of a line item (1 of 2 quantity) results in the correct reimbursement being generated.

Merge Checklist
---

- [x] Update the changelog


benjaminwil:
@forkata @Noah-Silvera Just want to update you on the status of this draft. (I will be away for the rest of the week.)

The commit "Fix tax calculation for orders with returned items" fixes the issue where bad reimbursements are generated during the RMA -> reimbursement admin flow. Check out the commit message for more info about 

But there is still something not right. If you examine the `refund.yml` cassette there are 400 responses coming back from the reporting API endpoints because the totals don't line up.

I haven't had time to delve in yet. I think the change Adam and I made is very correct but maybe something is still missing to account for the reporting API requests.

lib/super_good/solidus_taxjar/api_params.rb
```diff
@@ -111,46 +113,58 @@ def order_address_params(address)
...
+        # This method builds line item parameters as expected by the TaxJar
+        # Tax API.
+        #
+        # @param line_items [Spree::LineItem::ActiveRecord_Relation] All of the
+        #   order's line items.
+        # @return [Hash] A TaxJar API-friendly line item collection.
         def line_items_params(line_items)
           {
-            line_items: valid_line_items(line_items).map do |line_item|
+            line_items: line_items.filter_map { |line_item|
```
forkata, line 125: 💡 TIL `filter_map`!

spec/super_good/solidus_taxjar/api_params_spec.rb
```diff
@@ -221,27 +221,6 @@
       end
     end
 
-    context "when some line items have been returned" do
```
forkata, line 224: WDYT, about leave this test as a regression spec to ensure that returned/cancelled items are not excluded?

forkata (commented):
Great work @benjaminwil @adammathys!

spec/fixtures/cassettes/features/spree/admin/refund.yml
```diff
@@ -0,0 +1,572 @@
...
+  recorded_at: Tue, 23 Jan 2024 20:27:41 GMT
+- request:
+    method: post
+    uri: https://api.taxjar.com/v2/transactions/orders
+    body:
+      encoding: ASCII-8BIT
+      string: '{"customer_id":"1","to_country":"US","to_zip":"11430","to_city":"Herndon","to_state":"NY","to_street":"A
+        Different Road","line_items":[{"id":2,"quantity":1,"product_identifier":"SKU-2","description":"Second
+        Product - Master","product_tax_code":"TaxCode - 597842","unit_price":"10.0","discount":0,"sales_tax":"0.89"},{"id":3,"quantity":2,"product_identifier":"SKU-3","description":"Third
+        Product - Master","product_tax_code":"TaxCode - 597842","unit_price":"10.0","discount":0,"sales_tax":"1.78"}],"transaction_id":"R525233498","transaction_date":"2024-01-23T20:27:39Z","amount":"40.0","shipping":"0.0","sales_tax":"3.56"}'
```
forkata, line 314: It looks like we're not accounting for the reduced quantity in the `amount` field on the transaction here. This could be related to the fact that the event we hook into is fired before `persist_totals` is called, and thus giving us a false positive failure in the specs.

Noah-Silvera:
Just rebased to get the PR that adds a delay in here!

spec/features/spree/admin/refund_spec.rb
```diff
@@ -105,6 +105,9 @@
     # It will be -1 * a single inventory unit: $10 + the TaxJar-calculated tax
     # amount.
     #
+    if ENV["CI"]
+      pending "A very annoying and hard to debug issue is causing this check to fail consistently in CI, but not locally"
```
benjaminwil, line 109: If you care to tweak this... it is failing _regularly_ but not _consistently_. If you look at the CI runs it sometimes passes up to 2 of 4 Solidus versions.

Noah-Silvera: God that's so annoying!

forkata (approved):

PR #254, opened 2024-01-19
Get tests and binstubs working against Solidus `main`

forkata:
What is the goal of this PR?
---

1. To get the build against Solidus `main` green! This required to appease Zeitwerk by migrating one of our Deface overrides to the DSL format.
2. To be able to generate a sandbox app against Solidus `main`
3. To have a frontend to test with 😓 

Many thanks to @AlistairNorman for pairing on this with me as I was dreading having to work on this change!

How do you manually test these changes? (if applicable)
---

1. Visit the admin Taxes section
    * [x] Ensure the TaxJar settings tab is there
2. Test the `bin/sandbox` binstub on a few different branches, this was only tested on `main`

Merge Checklist
---

- [x] Run the manual tests
- [x] Update the changelog

Screenshots
---
N/A

PR #255, opened 2024-01-23
Add a delay before reporting transactions

benjaminwil:
What is the goal of this PR?
---

In a test environment, we discovered that sometimes these jobs would be run before an order's new totals had been persisted by the `Spree::OrderUpdater`. This is an effective workaround, but it is just a workaround.

Merge Checklist
---

- [x] Update the changelog


Noah-Silvera (approved):

forkata (approved):

forkata:
Looks like tests should pass here after we merge #254. Thanks for adding this @benjaminwil!

PR #260, opened 2026-09-16
Fix tests and errors

AlistairNorman:
What is the goal of this PR?
---

Fix all tests including an error caused by the require when using newer Rails versions. 

I don't love changing a migration but I think it makes sense in this case. Open to other suggestions on that one though

---

1. run tests/use the gem with Zeitwerk 

Merge Checklist
---

- [ ] Run the manual tests
- [ ] Update the changelog

Screenshots
---


adammathys (approved):

AlistairNorman:
ugh mysql causing issues again. I'll fix that up

forkata (approved):
Nice work @AlistairNorman!

jarednorman (approved):

